PR #31, opened 2019-05-10
Use correct month and year for rewards summary view.

iccub:


BraveRewards.xcodeproj/project.pbxproj
```diff
@@ -211,6 +215,16 @@
 /* End PBXContainerItemProxy section */
 
 /* Begin PBXCopyFilesBuildPhase section */
+		0A5AAECE2285E80B00F0EBC7 /* CopyFiles */ = {
+			isa = PBXCopyFilesBuildPhase;
+			buildActionMask = 2147483647;
+			dstPath = "";
+			dstSubfolderSpec = 10;
+			files = (
+				0A5AAED12285E81400F0EBC7 /* SnapKit.framework in CopyFiles */,
```
iccub: Had to add `snapkit`, otherwise couldn't import `BraveRewardsUI` for unit tests.
Found solution here https://github.com/Quick/Quick/issues/650#issuecomment-264332214

BraveRewardsUI/Supporting Files/Strings.swift
```diff
@@ -0,0 +1,37 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+
+struct Strings {
+  
+  struct DateTime {
```
iccub: I was thinking of splitting some strings in nested structs instead of just putting everything in one place, thoughts?

BraveRewardsUI/Extensions/DateExtensions.swift
```diff
@@ -0,0 +1,38 @@
...
+extension Date {
+  /// Returns localized name of current month, using Gregorian calendar format.
+  var currentMonthName: String {
+    let calendar = Calendar(identifier: .gregorian)
+    let monthNumber = calendar.component(.month, from: self)
+    
+    // Due to declensions, Apple's `DateFormatter.monthSymbols` can't be used to get localized month name.
+    // In Polish language for example, a different word variation is used depending on date format.
+    
+    switch monthNumber {
```
kylehickinson: Can't we just use iOS's Calendar strings (and just apply `uppercased`): https://developer.apple.com/documentation/foundation/calendar/2293077-monthsymbols

iccub: I thought we couldn't(see the comment lines above this line)
 but there's actually a `standaloneMonthSymbols` to help with this, fixing it

BraveRewardsTests/DateExtensionsTests.swift
```diff
@@ -0,0 +1,31 @@
...
+@testable import BraveRewardsUI
+
+class DateExtensionsTests: XCTestCase {
+
+  func testCurrentMonth() {
+    let february = dateFrom(string: "2018-02-22")
+    // Make sure tests are running on default english locale
+    XCTAssertEqual(february.currentMonthName, "February")
+  }
+
```
kylehickinson: Based on that previous comment about polish months: Is it possible add a locale specific test here too? :)  Not a problem if not, just as long as we test it manually

iccub: done

iccub: done

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -85,8 +85,10 @@ class WalletViewController: UIViewController {
       dollarValue: state.ledger.usdBalanceString
     )
     
+    let now = Date()
+    rewardsSummaryView.monthYearLabel.text = "\(now.currentMonthName().uppercased()) \(now.currentYear)"
+    
     // FIXME: Remove temp values
-    rewardsSummaryView.monthYearLabel.text = "MARCH 2019"
```
kylehickinson: Ah! Almost forgot there is a second usage of this 😆 Sorry!

https://github.com/brave/brave-rewards-ios/blob/a79aabdd8c878a89bfef03618c036b08b26b7c1d/BraveRewardsUI/Settings/Wallet/WalletDetailsViewController.swift#L44

iccub: Went with a slightly bigger change and put some shared code under a `protocol` to not miss it again. Lmk if that's too much of a refactor

BraveRewardsUI/Wallet/RewardsSummaryProtocol.swift
```diff
@@ -0,0 +1,44 @@
...
+  var state: RewardsState { get }
+  
+  /// Month and year of which the rewards summary is shown.
+  var summaryPeriod: String { get }
+  
+  /// Rows showing different types of earnings, tips etc.
+  var summaryRows: [RowView] { get }
+  
+  /// A view informing users about contributing to unverified publishers.
+  var disclaimerView: DisclaimerView { get }
```
kylehickinson: If we're going to put this in the protocol it has to be optional so it can be hidden when the user doesn't have auto-contribute enabled or doesn't have any auto-contribution entries.  Maybe we should get clarification from @NejcZdovc or just double check in brave-core when that disclaimer is supposed to be shown.

BraveRewardsUI/Wallet/RewardsSummaryProtocol.swift
```diff
@@ -0,0 +1,44 @@
...
+
+/// Shared resources for showing summary of all BAT rewards.
+protocol RewardsSummaryProtocol {
+  var state: RewardsState { get }
+  
+  /// Month and year of which the rewards summary is shown.
+  var summaryPeriod: String { get }
+  
+  /// Rows showing different types of earnings, tips etc.
+  var summaryRows: [RowView] { get }
```
kylehickinson: This is fine for now, but may need modification in the future :)

PR #33, opened 2019-05-13
Replace fake rewards summary values with real values.

iccub:


kylehickinson:
Nice!! Maybe we can get 1 thing in before merging this too.  

There's actually supposed to be colors associated with these rewards summary values as can be seen here: 

Untitled

I asked James what these colors meant since they dont belong to the color palette and they are the same colors as used in the settings screen for Ads/Auto-Contribute/Tips

Basically: 
- Grants+Ads = Ads title colour
- Auto-Contribution = AC title colour (Purple)
- Tips = Tips title color (blurple)

Can we get these colors added into the RowView's in this PR since they are related to that view's integration?

iccub:
Sounds good to me, i'll update it once James will provide color values

kylehickinson:
The colours should be here: 

https://github.com/brave/brave-rewards-ios/blob/c576f192c84e68f69d58ad94f3d4dbb83a28760d/BraveRewardsUI/Common/Colors.swift#L44-L46 

Under `BraveUX`

iccub:
I misread your comment, sorry.

Updated colors:

Zrzut ekranu 2019-05-13 o 21 30 29

BraveRewardsUI/Extensions/NumberExtensions.swift
```diff
@@ -0,0 +1,13 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+
+extension Double {
+  /// Format of which BAT should be presented to user.
+  /// Similar method also exists for Strings.
+  var BATDisplayFormat: String {
```
kylehickinson: Slight naming change I think here. It sounds like this should be returning some String format (like `"%.1"`) and not the actual BAT string.  Perhaps something like `formattedBATDisplayString`?

It actually makes me wonder if we should just make a `BATValue` struct or something that takes a double/string/probi and lets us pull out whatever type we want?  Then we won't have to make extensions straight on `String`/`Double`

For grants we're actually going to get a `String` but it won't be in the format of like "30.0", it'll be a probi string which will be `"30000000000000000000"` (30 * 10^18)

Also lets us isolate tests around BAT values to one struct which is a plus. 

Let me know what you think about that.

iccub: I like it, especially if there's more than just simple String and Double formats coming.
I will refactor this tomorrow

kylehickinson:
Example of probi string from grants: https://github.com/brave/brave-core/blob/master/test/data/rewards-data/captcha_solution_resp.json

Issue #19, opened 2019-04-24
Add Strings.swift and replace usages of BATLocalizedString


kylehickinson:
Currently all user-facing strings are wrapped with [`BATLocalizedString`](https://github.com/brave/brave-rewards-ios/blob/7036e865e4727815510cf129b2e45f5ef905d8de/BraveRewardsUI/Extensions/BundleExtensions.swift#L9).

We should switch to the same system as `brave-ios`, where we use a `Strings.swift` file containing static accessors to all localized strings

Issue #20, opened 2019-04-24
Add localization scripts


kylehickinson:
Port over the localization scripts [found in `brave-ios`](https://github.com/brave/brave-ios/tree/development/l10n) used to push/pull strings from transifex

PR #52, opened 2019-05-16
Fixed: #19 & #20 Added Localisation support for Brave Rewards

danishjafri88:
Ported l10n script from Brave-iOS repo
Created a Strings struct to hold all localisable strings.
Pushed xliff to transifex

Merging two issues as they are related.


kylehickinson:
Does this merge any strings I may have duped?

PR #61, opened 2019-05-20
Publisher database unit tests

iccub:


BraveRewards/Ledger/Data/BATLedgerDatabase.mm
```diff
@@ -438,13 +438,18 @@ + (void)insertOrUpdateRecurringTipWithPublisherID:(NSString *)publisherID
 
 + (BOOL)removeRecurringTipWithPublisherID:(NSString *)publisherID
 {
-  const auto context = [DataController newBackgroundContext];
-  const auto rd = [self getRecurringDonationWithPublisherID:publisherID context:context];
-  if (!rd) {
+  // Early guard to check if object exists.
+  // The check happens on main thread, while the deletion is done in background.
+  const auto donationExists = [self getRecurringDonationWithPublisherID:publisherID
```
iccub: previous code caused threading conflicts, made a small change

BraveRewards/Ledger/Data/BATLedgerDatabase.mm
```diff
@@ -291,7 +291,7 @@ + (void)insertOrUpdateActivitiesInfoFromPublishers:(NSArray<BATPublisherInfo *>
   }
   
   if (filter.minDuration > 0) {
-    [predicates addObject:[NSPredicate predicateWithFormat:@"duration >= %ld", filter.reconcileStamp]];
+    [predicates addObject:[NSPredicate predicateWithFormat:@"duration >= %ld", filter.minDuration]];
```
iccub: that was a wrong predicate argument, I guess copy/paste bug

BraveRewards/Ledger/Data/BATLedgerDatabase.mm
```diff
@@ -438,13 +438,18 @@ + (void)insertOrUpdateRecurringTipWithPublisherID:(NSString *)publisherID
...
-  [DataController.shared performOnContext:context task:^(NSManagedObjectContext * _Nonnull context) {
-    [context deleteObject:rd];
+  
+  [DataController.shared performOnContext:nil task:^(NSManagedObjectContext * _Nonnull context) {
+    const auto donation = [self getRecurringDonationWithPublisherID:publisherID context:context];
+    
+    [context deleteObject:donation];
     // Probably want to adjust `performOnContext` to call this in a scenario where we pass an
     // existing context
     if (context.hasChanges) {
```
kylehickinson: We don't need this manual save anymore now that we are not executing with a custom context

BraveRewardsTests/LedgerDatabaseTests.mm
```diff
@@ -144,6 +147,638 @@ - (void)testNonZeroExcludedCount
...
+  info.percent = percent;
+  info.duration = duration;
+  info.visits = visits;
+  info.excluded = excluded;
+  
+  [self backgroundSaveAndWaitForExpectation:^{
+    [BATLedgerDatabase insertOrUpdatePublisherInfo:info];
+  }];
+  
+  if(createActivityInfo){
```
kylehickinson: [Minor/Linting]: Spaces around brackets like `if (createActivityInfo) {`

BraveRewardsTests/LedgerDatabaseTests.mm
```diff
@@ -144,6 +147,638 @@ - (void)testNonZeroExcludedCount
...
+    [self backgroundSaveAndWaitForExpectation:^{
+      [BATLedgerDatabase insertOrUpdateActivityInfoFromPublisher:info];
+    }];
+  }
+  
+  return info;
+}
+
+- (void) testInsertOrUpdatePublisherInfoEmptyId
```
kylehickinson: [Minor/Linting]: No space after bracket like `- (void)testInsert…`

BraveRewardsTests/LedgerDatabaseTests.mm
```diff
@@ -144,6 +147,638 @@ - (void)testNonZeroExcludedCount
...
+  [self makeOneTimeTip:@"444" month:BATActivityMonthJanuary year:year];
+  
+  const auto newerTips = [BATLedgerDatabase oneTimeTipsPublishersForMonth:month year:year];
+  XCTAssert([self countMustBeEqualTo:6 forEntityName:@"ContributionInfo"]);
+  XCTAssertEqual(newerTips.count, 3);
+}
+
+- (void)makeOneTimeTip:(NSString *)publisherId month:(const int)month year:(const int)year
+{
+  const auto probi = @"10000000000000000";
```
kylehickinson: Since this method is a duplicate of the one below minus some arguments, it should call the other with said default arguments.

Example:

```objc
- (void)makeOneTimeTip:(NSString *)publisherId month:(const int)month year:(const int)year
{
  [self makeOneTimeTip:publisherId month:month year:year probi:@"10000000000000000" category:BATRewardsCategoryOneTimeTip];
}
```

BraveRewardsTests/LedgerDatabaseTests.mm
```diff
@@ -144,6 +147,638 @@ - (void)testNonZeroExcludedCount
...
+  [self createBATPublisherInfo:@"1" reconcileStamp:10 percent:11 createActivityInfo:YES];
+  [self createBATPublisherInfo:@"2" reconcileStamp:10
+                                               percent:22 createActivityInfo:YES];
+  
+  XCTAssert([self countMustBeEqualTo:2 forEntityName:@"ActivityInfo"]);
+  
+  [self backgroundSaveAndWaitForExpectation:^{
+    [BATLedgerDatabase deleteActivityInfoWithPublisherID:@"1" reconcileStamp:10];
+  }];
+  XCTAssert([self countMustBeEqualTo:1 forEntityName:@"ActivityInfo"]);
```
kylehickinson: Probably doesn't really matter, but do you think we should query the remaining activity info's to verify 2/10/22 is still there? All we verify by checking count is that we deleted 1 object, but didn't check which object was deleted.

BraveRewardsTests/LedgerDatabaseTests.mm
```diff
@@ -144,6 +147,638 @@ - (void)testNonZeroExcludedCount
...
+  }];
+  
+  XCTAssert([self countMustBeEqualTo:3 forEntityName:@"ActivityInfo"]);
+}
+
+- (void)testDeleteActivityInfoWithPublisherID
+{
+  [self createBATPublisherInfo:@"1" reconcileStamp:10 percent:11 createActivityInfo:YES];
+  [self createBATPublisherInfo:@"2" reconcileStamp:10
+                                               percent:22 createActivityInfo:YES];
```
kylehickinson: [Minor/Lint]: New line here probably isn't needed?

BraveRewardsTests/LedgerDatabaseTests.mm
```diff
@@ -144,6 +147,638 @@ - (void)testNonZeroExcludedCount
...
+
+#pragma mark - publishersWithActivityFromOffset
+
+- (void)testPublishersWithFiltersEmptyFilter
+{
+  const auto offset = 0;
+  const auto limit = 0;
+  const auto filter = [[BATActivityInfoFilter alloc] init];
+  
+  [self createBATPublisherInfo:@"1" reconcileStamp:10 percent:10 createActivityInfo:true];
```
kylehickinson: [Minor/Lint]: I see you started using `true` in some places, and `YES` in others. It doesn't matter since they will compile to the same thing, but for consistency sake I'd rather we stick with one. I've used `YES` throughout, out of Obj-C habit, but if you want to use `true` instead to match Swift for this file thats fine by me.

BraveRewardsTests/LedgerDatabaseTests.mm
```diff
@@ -144,6 +147,638 @@ - (void)testNonZeroExcludedCount
...
+  [self createBATPublisherInfo:@"2" reconcileStamp:10 percent:10 createActivityInfo:true];
+  [self createBATPublisherInfo:@"3" reconcileStamp:10 percent:10 createActivityInfo:true];
+  
+  const auto limit = 1;
+  const auto resultWithLimit = [BATLedgerDatabase publishersWithActivityFromOffset:0 limit:limit filter:filter];
+  
+  XCTAssertEqual(resultWithLimit.count, 1);
```
kylehickinson: Same comment as above in regards to testing only count instead of result of the action (in this case, querying an item).  Maybe intended based on the test name?

BraveRewardsTests/LedgerDatabaseTests.mm
```diff
@@ -144,6 +147,638 @@ - (void)testNonZeroExcludedCount
...
+  const auto expectedIds = @[@"3", @"5", @"4"];
+  
+  // Make sure records by `reconcileStamp` are in wrong order when using one order pair
+  XCTAssertFalse([oneSortResultIds isEqualToArray:expectedIds]);
+  
+  const auto stampAscending = [[BATActivityInfoFilterOrderPair alloc] init];
+  stampAscending.propertyName = @"reconcileStamp";
+  [stampAscending setAscending:YES];
+  
+  [filter setOrderBy:@[percentAscending, stampAscending]];
```
kylehickinson: [Minor/Lang]: You can call setters using dot notation, same as Swift like `stampAscending.ascending = YES;` and `filter.orderBy = @[percentAscending, stampAscending];`

Is there a reason you called the `setX` method in these few places here and above?

BraveRewardsTests/LedgerDatabaseTests.mm
```diff
@@ -144,6 +147,638 @@ - (void)testNonZeroExcludedCount
...
+#pragma mark - Media Publisher Info
+
+- (void)testInsertOrUpdateMediaPublisherInfoWithMediaKey
+{
+  [self createBATPublisherInfo:@"1" reconcileStamp:40 percent:0 createActivityInfo:true];
+  
+  XCTAssert([self countMustBeEqualTo:0 forEntityName:@"MediaPublisherInfo"]);
+  
+  [self waitForFailedSaveAttempt:^{
+  [BATLedgerDatabase insertOrUpdateMediaPublisherInfoWithMediaKey:@"" publisherID:@"1"];
```
kylehickinson: Minor/Lint: Spacing before this call. I assume the wait and fail was just added after :)

BraveRewardsTests/LedgerDatabaseTests.mm
```diff
@@ -144,6 +147,638 @@ - (void)testNonZeroExcludedCount
...
+  
+  const auto date = [[dateFormatter dateFromString:@"2019-10-07"] timeIntervalSince1970];
+  const auto date2 = [[dateFormatter dateFromString:@"2020-11-17"] timeIntervalSince1970];
+  
+  [self createBATPublisherInfo:@"1" reconcileStamp:40 percent:0 createActivityInfo:true];
+  
+  [self createRecurringTip:publisherId amount:15.0 date:date];
+  [self createRecurringTip:@"2" amount:100.0 date:date2];
+  
+  XCTAssert([self countMustBeEqualTo:2 forEntityName:@"RecurringDonation"]);
```
kylehickinson: Is there a reason to do both the DataController count assertion and the object query count assertion below?

BraveRewardsTests/LedgerDatabaseTests.mm
```diff
@@ -196,4 +831,18 @@ - (void)backgroundSaveAndWaitForExpectation:(void (^)())task
   [self waitForExpectations:@[saveExpectation] timeout:5];
 }
 
+/// Waits a second to verify that no save happened.
+// Useful when you want to test saves with wrong values, early returns etc.
```
kylehickinson: Minor: Missing the 3rd `/` here 😅

kylehickinson:
Mostly minor/linting stuff. 

One big thing I saw though was a lot of count testing over actual result testing. I would expect that testing the objects queried or inserted would be the better test, but let me know if you don't think theres much value for these tests for some reason

BraveRewardsTests/LedgerDatabaseTests.mm
iccub: We can't access database object from tests, I haven't found solution to it yet

iccub:
I would love to test object's contents for tests but we can't access them at the moment

kylehickinson:
As discussed: Will use `BATLedgerDatabase` getter methods and check the returned `BATPublisherInfo`(s) for the info needed instead of obtaining info directly from DB objects

BraveRewardsTests/LedgerDatabaseTests.mm
iccub: i remember that Xcode complained something about it and suggested to use `setX` instead, 
it's just me fighting with objc syntax, changing it to dot notation

BraveRewardsTests/LedgerDatabaseTests.mm
```diff
@@ -144,6 +147,663 @@ - (void)testNonZeroExcludedCount
...
+  // Make sure correct activity got deleted and remaning one still exists.
+  const auto filter = [[BATActivityInfoFilter alloc] init];
+  filter.id = pubId;
+  
+  auto result = [BATLedgerDatabase publishersWithActivityFromOffset:0 limit:0 filter:filter];
+  XCTAssertEqual(result.count, 0);
+  
+  filter.id = @"2";
+  result = [BATLedgerDatabase publishersWithActivityFromOffset:0 limit:0 filter:filter];
+  XCTAssertEqual(result.count, 1);
```
kylehickinson: Why bother checking the count here for 1? Why not just check that the object?

PR #73, opened 2019-05-28
Get publisher by url, set verification badge and attention value.

iccub:


BraveRewards/BATBraveRewards.m
```diff
@@ -100,6 +100,7 @@ - (void)setupLedgerAndAds
   NSString *ledgerStorage = [self.configuration.stateStoragePath stringByAppendingPathComponent:@"ledger"];
   self.ledger = [[self.ledgerClass alloc] initWithStateStoragePath:ledgerStorage];
   self.ledger.ads = self.ads;
+  [[self ledger] insertTestPublishers];
```
kylehickinson: Shouldn't be called here, probably need a better way of adding test info (i.e. calling reporting methods with a button in example app)

BraveRewards/Ledger/BATBraveLedger.h
```diff
@@ -212,6 +215,10 @@ NS_SWIFT_NAME(BraveLedger)
 /// Get the number of ads received and the estimated earnings of viewing said ads for this cycle
 - (void)adsDetailsForCurrentCycle:(void (^)(NSInteger adsReceived, double estimatedEarnings))completion NS_SWIFT_NAME(adsDetailsForCurrentCycle(_:));
 
+#pragma mark - Test only
+
+- (void)insertTestPublishers;
```
kylehickinson: Should be in a private category whose header access level is `Project`

BraveRewards/Ledger/BATBraveLedger.mm
```diff
@@ -1117,4 +1127,22 @@ - (void)savePublisherInfo:(std::unique_ptr<ledger::PublisherInfo>)publisher_info
   }
 }
 
+#pragma mark - Test only
+
+// This is temporary until we integrate brave-rewards with brave-ios.
+- (void)insertTestPublishers
```
kylehickinson: Same as above, private category

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -152,9 +152,18 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
...
       
-      // FIXME: Remove fake data
-      publisherView.setVerified(true)
-      attentionView.valueLabel.text = "19%"
+      guard let host = state.url.host else { return }
+      
+      state.ledger.publisherInfo(forId: host) { info in
+        guard let publisher = info else { return }
+        
+        DispatchQueue.main.async {
```
kylehickinson: We should probably jump to main from the actual ledger method call site instead of at UI level (should technically always execute completion blocks on main unless manually specified)

BraveRewards/Ledger/BATBraveLedger.h
```diff
@@ -88,17 +88,19 @@ NS_SWIFT_NAME(BraveLedger)
 - (void)activityInfoWithFilter:(nullable BATActivityInfoFilter *)filter
                     completion:(void (^)(BATPublisherInfo * _Nullable info))completion;
 
-// Unavailable until we figure out what it actually does and how to use it
 - (void)publisherActivityFromURL:(NSURL *)URL
                       faviconURL:(NSURL *)faviconURL
                         windowID:(uint64_t)windowID
```
kylehickinson: We should just remove this argument from the public API, since I thought we were always going to send in 1 to the actual ledger API?

iccub:
This is how the panel looks without AC setting enabled btw

Zrzut ekranu 2019-05-30 o 23 22 41

and on desktop

Zrzut ekranu 2019-05-30 o 23 26 06

Issue #66, opened 2019-05-21
Add support for badges on the BraveRewards/BAT URL bar button


kylehickinson:
Verified publishers show a checkmark in the Brave Rewards button in the URL bar.

Considering that we also have to show a badge count for notifications, it may be better to create a custom button that handles badges on the icon that can just be added directly in brave-ios

kylehickinson:
This will happen in https://github.com/brave/brave-ios/pull/1169

Issue #45, opened 2019-05-15
Show notifications on the main Rewards panel when available


kylehickinson:
When the user has notifications to view, they should be displayed on `WalletViewController` until dismissed manually (or possibly time based?) 

These notifications may have actions (like #44 for claiming a grant) which can be integrated in different tickets

~Blocked by #41~

PR #75, opened 2019-05-29
Fix #41: Add notifications implementation

### LINKED ISSUE #41: Rewards Notifications Timers + Implementations
Set up notification polling timers following same implementation as brave-core:

See notification methods in [`rewards_service_impl.cc`](https://github.com/brave/brave-core/blob/de608a5c617b9439e5a835ba575d7b12ae8b7d42/components/brave_rewards/browser/rewards_service_impl.cc#L827)

Each timer hit checks a few different spots to determine if a notification should be added. Said notifications should be stored somewhere in the `BATBraveLedger` for the UI to pickup and use.

Likely we will have to add a delegate to the `BATBraveLedger` that will notify the UI of many changes (publishers added, notifications shown, etc.). UI will likely forward some of those to client (notification badge, fetch favicon, etc.)

kylehickinson:


BraveRewards/Ledger/BATBraveLedger.mm
```diff
@@ -72,7 +86,20 @@ - (instancetype)initWithStateStoragePath:(NSString *)path
     self.storagePath = path;
     self.commonOps = [[BATCommonOperations alloc] initWithStoragePath:path];
     self.state = [[NSMutableDictionary alloc] initWithContentsOfFile:self.randomStatePath] ?: [[NSMutableDictionary alloc] init];
-    self.stateWriteThread = dispatch_queue_create("com.rewards.state", DISPATCH_QUEUE_SERIAL);
+    self.fileWriteThread = dispatch_queue_create("com.rewards.state", DISPATCH_QUEUE_SERIAL);
```
iccub: Change dispatchqueues name to match the renamed `fileWriteThread`

BraveRewards/Ledger/BATBraveLedger.mm
```diff
@@ -452,6 +497,16 @@ - (void)onGrant:(ledger::Result)result grant:(const ledger::Grant &)grant
 {
   if (result == ledger::LEDGER_OK) {
     [self.mPendingGrants addObject:[[BATGrant alloc] initWithGrant:grant]];
+    
+    bool isUGP = grant.type == "ugp";
```
iccub: Isn't there a constant for `"ugp"` type? 
If not, I would extract it to constant somewhere, we use it in two places

BraveRewards/Ledger/BATBraveLedger.mm
```diff
@@ -452,6 +497,16 @@ - (void)onGrant:(ledger::Result)result grant:(const ledger::Grant &)grant
...
+    auto notificationKind = isUGP ? BATRewardsNotificationKindGrant : BATRewardsNotificationKindGrantAds;
+    auto prefix = isUGP ? @"rewards_notification_grant_" : @"rewards_notification_grant_ads_";
```
iccub: These strings could also be extracted somewhere, also used in few places

BraveRewards/Ledger/BATBraveLedger.mm
```diff
@@ -552,6 +613,22 @@ - (void)onReconcileComplete:(ledger::Result)result viewingId:(const std::string
                                        nowTimestamp);
   }
   
+  if ((result == ledger::Result::LEDGER_OK && category == ledger::REWARDS_CATEGORY::AUTO_CONTRIBUTE) ||
+      result == ledger::Result::LEDGER_ERROR ||
```
iccub: Do we want to add notification even if `result` has error states?

BraveRewards/Ledger/BATBraveLedger.mm
```diff
@@ -757,7 +834,205 @@ - (void)confirmationsTransactionHistoryDidChange
...
+  self.notificationStartupTimer =
+  [NSTimer scheduledTimerWithTimeInterval:30
+                                   target:self
+                                 selector:@selector(notificationTimerFired)
+                                 userInfo:nil
+                                  repeats:NO];
+  
+  // Periodic timer, runs once per day by default.
+  self.notificationCheckTimer =
+  [NSTimer scheduledTimerWithTimeInterval:60 * 60 * 24
```
iccub: 24 hours is a lot and the app may get killed by that time, I guess this is just a fallback timer.
And usually when user will open the app on next day `notificationStartupTimer` will launch after 30 seconds, right?

BraveRewards/Ledger/BATBraveLedger.mm
```diff
@@ -757,7 +834,205 @@ - (void)confirmationsTransactionHistoryDidChange
...
+                                 selector:@selector(notificationTimerFired)
+                                 userInfo:nil
+                                  repeats:YES];
+}
+
+- (void)notificationTimerFired
+{
+  [self showBackupNotificationIfNeccessary];
+  [self showAddFundsNotificationIfNeccessary];
+  [self fetchGrants:std::string() paymentId:std::string()];
```
iccub: Why fetching grants here?

BraveRewards/Ledger/BATBraveLedger.mm
```diff
@@ -757,7 +834,205 @@ - (void)confirmationsTransactionHistoryDidChange
...
+  
+  const auto notification = [[BATRewardsNotification alloc] initWithID:notificationID
+                                                             dateAdded:[[NSDate date] timeIntervalSince1970]
+                                                                  kind:kind
+                                                             arguments:arguments];
+  if (onlyOnce) {
+    notification.displayed = YES;
+  }
+  
+  [self.mNotifications addObject:notification];
```
iccub: These notifications never launch, only `BATBraveLedgerNotificationAdded` is posted, are you going to add it in another PR or here? Maybe i'm missing something

BraveRewards/Ledger/BATBraveLedger.mm
kylehickinson: Logic was copied from brave-core's implementation

BraveRewards/Ledger/BATBraveLedger.mm
kylehickinson: Hmm, the logic was copied from brave-core's implementation but yeah perhaps we should change ours based on background/foregrounding?

But yeah, if the app is ejected from memory and relaunches it'll check again in 30s

BraveRewards/Ledger/BATBraveLedger.mm
kylehickinson: Copied from brave-core's implementation

BraveRewards/Ledger/BATBraveLedger.mm
kylehickinson: These notifications won't show up as regular apple notifications, they'll update the badge (#66),  and show new UI on the panel when the user next shows it (#45)

BraveRewards/Ledger/BATBraveLedger.mm
kylehickinson: There's no constant in the ledger lib, but I can extract them at a least here

BraveRewards/Ledger/BATBraveLedger.mm
kylehickinson: Yeah, so as seen here: https://github.com/brave/brave-core/blob/f6956d15bbc9ccf51faceafe5d74c4c3f4e525b2/components/brave_rewards/browser/rewards_notification_service_impl.cc#L348

The result is passed into the notification info, so I assume when we go to implement the UI shown for this notification (#45), it will change based on what the result here is. Since this is actually going to happen in the UI, I'm going to change this to not be a string

BraveRewards/Ledger/BATBraveLedger.mm
```diff
@@ -757,7 +845,199 @@ - (void)confirmationsTransactionHistoryDidChange
...
+}
+
+- (void)showAddFundsNotificationIfNeccessary
+{
+  const auto stamp = ledger->GetReconcileStamp();
+  const auto now = [[NSDate date] timeIntervalSince1970];
+  
+  // Show add funds notification if reconciliation will occur in the
+  // next 3 days and balance is too low.
+  if (stamp - now > 3 * 24 * 60) {
```
iccub: missing one more `60` for minutes here, can make a constant for it, as it's used in two places in this method

BraveRewards/Ledger/BATBraveLedger.mm
```diff
@@ -34,6 +34,13 @@ + (__type)__objc_getter { return ledger::__cpp_var; } \
 + (void)__objc_setter:(__type)newValue { ledger::__cpp_var = newValue; }
 
 NSString * const BATBraveLedgerErrorDomain = @"BATBraveLedgerErrorDomain";
+NSNotificationName const BATBraveLedgerNotificationAdded = @"BATBraveLedgerNotificationAdded";
+
+static NSString * const kNextAddFundsNotificationKey = @"BATNextAddFundsNotification";
```
iccub: rename to `kNextAddFundsDateNotificationKey` to make it clear that it is a date we store here

PR #82, opened 2019-06-04
Complete move of BraveRewards to brave-core

kylehickinson:
- Removes BraveRewards framework target
- Update ledger implementation based on newest core changes
- Update build script that builds from core
- Adds BraveRewardsCore aggregate target. Only works assuming the location of your brave-browser is "$(PROJECT_DIR)/../brave-browser"
- Now pushes FAT compiled framework to the repo
- Updates README

Also:
- Adds SwiftLint
- Adds Travis CI


.swiftlint.yml
```diff
@@ -0,0 +1,73 @@
+disabled_rules: # rule identifiers to exclude from running
```
iccub: Is it the same list as in brave-ios?

kylehickinson: Yep, exact same swiftlint rules

PR #100, opened 2019-06-18
Add ability to set background color via retrieveFavicon.

iccub:
This is needed so we can set favicons correctly, for predefined websites and for custom letter+background websites.

### DIFF
--- a/BraveRewardsUI/RewardsDataSource.swift
+++ b/BraveRewardsUI/RewardsDataSource.swift
@@ -13,6 +13,6 @@ public protocol RewardsDataSource: AnyObject {
   /// Download or retrieve a cached version of the favicon given a URL. This should also return the
   /// default letter/color favicon if a website has no official favicon.
   ///
-  /// Execute completionBlock with an image (or nil) when its available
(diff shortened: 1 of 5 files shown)


BraveRewardsUI/RewardsDataSource.swift
```diff
@@ -13,6 +13,6 @@ public protocol RewardsDataSource: AnyObject {
   /// Download or retrieve a cached version of the favicon given a URL. This should also return the
   /// default letter/color favicon if a website has no official favicon.
   ///
-  /// Execute completionBlock with an image (or nil) when its available
-  func retrieveFavicon(with url: URL, completion: @escaping (UIImage?) -> Void)
+  /// Execute completionBlock with an image and background color (or nil) when its available
+  func retrieveFavicon(with url: URL, completion: @escaping (UIImage?, UIColor?) -> Void)
```
kylehickinson: Is there ever a point where we'd send non-nil image and nil color (or vice-versa)? 

If not, we should probably just package this up into a struct, like `Favicon` or an optional tuple.

PR #101, opened 2019-06-20
Fix #48: Integrated Total tips, and tip contributions in Tips details

### LINKED ISSUE #48: Integrate Tips Detail & List
In settings, in the tips section, you can tap the View Details button. On this new screen, we should:

- [x] Total tips this month value should be computed correctly
- [x] Combine two data points: one-time tips (`listOneTimeTips:`), recurring tips (`listRecurringTips:`) and 
- [x] Media publishers may show a name + `on YouTube`/`on Twitch` instead of domain
- [x] Recurring Tips are editable by swipe-to-"remove" (cancelling the recurring payment)
- [x] Recurring Tips subtitle is "Recurring"
- [x] One-Time Tips subtitle is "One time" and shows a date next to it
- [ ] If the user has any recurring tips, a new cell appears that looks the same as Auto-Contribute/Ads details where it shows the next monthly payment date

Simulator Screen Shot - iPhone X - 2019-05-15 at 17 00 58

Desktop ref: https://brave.github.io/brave-ui/?path=/story/feature-components-rewards-table--donation

danishjafri88:


BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
```diff
@@ -112,35 +140,62 @@ extension TipsDetailViewController: UITableViewDataSource, UITableViewDelegate {
...
+      
       cell.siteImageView.image = UIImage(frameworkResourceNamed: "defaultFavicon")
-      cell.verifiedStatusImageView.isHidden = indexPath.row != 0
-      cell.typeNameLabel.text = "Recurring"
+      setFavicon(url: tip.faviconUrl, indexPath: indexPath)
+      cell.verifiedStatusImageView.isHidden = !tip.verified
+      switch tip.category {
+      case .oneTimeTip:
+        // FIXME: Validate if the reconcileStamp is the right property to check last tip time.
+        cell.typeNameLabel.text = "One time " + Date.dateStringFrom(tip.reconcileStamp)
```
danishjafri88: Are these supposed to be localised?

kylehickinson: Yeah I assume so

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
```diff
@@ -36,6 +39,32 @@ class TipsDetailViewController: UIViewController {
...
+      totalBatTips = $0.total.isEmpty ? "0.0" : $0.total
+    }
+    
+    let dateFormatter = DateFormatter().then {
+      $0.dateStyle = .short
+      $0.timeStyle = .none
+    }
+    // Fixme: Check if recurreing tips happen at autoContribute reconcileStamp
+    let reconcileDate = Date(timeIntervalSince1970: TimeInterval(state.ledger.autoContributeProps.reconcileStamp))
+    nextContributionDateView.label.text = dateFormatter.string(from: reconcileDate)
```
kylehickinson: Can't you use the `dateStringFrom ` method you added?

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
```diff
@@ -36,6 +39,32 @@ class TipsDetailViewController: UIViewController {
...
+    nextContributionDateView.bounds = CGRect(origin: .zero, size: nextContributionDateView.systemLayoutSizeFitting(UIView.layoutFittingExpandedSize))
+    
+    state.ledger.listOneTimeTips {[weak self] infoList in
+      self?.tipsList.append(contentsOf: infoList)
```
kylehickinson: Should `guard self = self else { return }` if you're going to use weak self more than once

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
```diff
@@ -36,6 +39,32 @@ class TipsDetailViewController: UIViewController {
...
+      (self?.view as? View)?.tableView.reloadData()
+    }
+    
+    state.ledger.listRecurringTips {[weak self] infoList in
+      self?.tipsList.insert(contentsOf: infoList, at: 0)
```
kylehickinson: Same as above

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
```diff
@@ -94,15 +123,18 @@ extension TipsDetailViewController: UITableViewDataSource, UITableViewDelegate {
     guard let typedSection = Section(rawValue: section) else { return 0 }
     switch typedSection {
     case .summary:
-      return 1
+      return 1 + (tipsList.filter({$0.category == .recurringTip}).isEmpty ? 0 : 1 )
```
kylehickinson: lint/format: un-needed enclosing parenthesis

BraveRewardsUI/Extensions/DateExtensions.swift
```diff
@@ -27,4 +27,13 @@ extension Date {
     let calendar = Calendar(identifier: .gregorian)
     return calendar.component(.month, from: self)
   }
+  
+  static func dateStringFrom(_ intervalSince1970: UInt64) -> String {
```
kylehickinson: API naming: `dateStringFrom(_:)` → `dateString(from:)` or `dateStringFrom(intervalSince1970:)`

kylehickinson: Required tests: Other method in this extension are tested in [`DateExtensionTests`](https://github.com/brave/brave-rewards-ios/blob/master/BraveRewardsTests/DateExtensionsTests.swift).  Should add a test here to test this method

kylehickinson: Also why is `intervalSince1970` a `UInt64` instead of a `TimeInterval`?

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
```diff
@@ -11,10 +11,13 @@ class TipsDetailViewController: UIViewController {
     return view as! View // swiftlint:disable:this force_cast
   }
   
-  private let ledger: BraveLedger
+  private let nextContributionDateView =  LabelAccessoryView()
```
kylehickinson: lint/format: extra space after =

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
```diff
@@ -112,35 +144,61 @@ extension TipsDetailViewController: UITableViewDataSource, UITableViewDelegate {
...
+        let cell = tableView.dequeueReusableCell(for: indexPath) as EmptyTableCell
+        cell.label.text = Strings.EmptyTipsText
+        return cell
+      }
       let cell = tableView.dequeueReusableCell(for: indexPath) as TipsTableCell
-      cell.siteNameLabel.text = "theguardian.com"
+      let tip = tipsList[indexPath.row]
+      // Fixme: Check if provider name requries change in text color. If so cahnge to usign attributed string.
+      let provider = "\(tip.provider.isEmpty ? "" : "on \(tip.provider)")"
+      cell.siteNameLabel.text = "\(tip.name) \(provider)"
```
kylehickinson: I believe the provider is supposed to styled differently (bold?) in design

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
```diff
@@ -182,3 +247,24 @@ extension TipsDetailViewController {
...
+
+extension TipsDetailViewController {
+  fileprivate func getTipsThisMonth() -> BalanceReportInfo {
+    let components = Calendar(identifier: .gregorian).dateComponents([.month, .year], from: Date())
+    let month = components.month ?? -1
+    let year = components.year ?? 2019
+    return state.ledger.balanceReport(for: ActivityMonth(rawValue: month) ?? .any, year: Int32(year))
+  }
+  
+  fileprivate func setFavicon(url: String, indexPath: IndexPath) {
```
kylehickinson: Using index path here is not a good way of updating the cell, since index paths could change if the table is reloaded at any point.

Better to pass in the tip, which then give you an identifier to update based on where the tip is in the list rather than where it was when you started the request.

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
```diff
@@ -182,3 +247,24 @@ extension TipsDetailViewController {
...
+    if let faviconURL = URL(string: url) {
+      state.dataSource?.retrieveFavicon(with: faviconURL, completion: {[weak self] favData in
+        if let tableView = (self?.view as? View)?.tableView,
+          tableView.indexPathsForVisibleRows?.contains(indexPath) == true,
+          let cell = tableView.cellForRow(at: indexPath) as? TipsTableCell {
+          cell.siteImageView.image = favData?.image ?? UIImage(frameworkResourceNamed: "defaultFavicon")
```
kylehickinson: You should check if `favData` is non-nil first and just return early if its nil (since you already set it as defaultFavicon at cell setup)

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
```diff
@@ -182,3 +247,24 @@ extension TipsDetailViewController {
     }
   }
 }
+
+extension TipsDetailViewController {
+  fileprivate func getTipsThisMonth() -> BalanceReportInfo {
+    let components = Calendar(identifier: .gregorian).dateComponents([.month, .year], from: Date())
+    let month = components.month ?? -1
+    let year = components.year ?? 2019
```
kylehickinson: Don't like defaulting to any given year, should just change this to return `BalanceReportInfo?` and return nil if any info is missing, making it clear we're not receiving a valid balance report instead of just getting an empty one

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
```diff
@@ -149,8 +207,15 @@ extension TipsDetailViewController: UITableViewDataSource, UITableViewDelegate {
...
   func tableView(_ tableView: UITableView, commit editingStyle: UITableViewCell.EditingStyle, forRowAt indexPath: IndexPath) {
-    guard Section(rawValue: indexPath.section) == .tips else { return }
-    
+    guard Section(rawValue: indexPath.section) == .tips &&
+      !tipsList.isEmpty &&
+      tipsList[indexPath.row].category == .recurringTip else { return }
+    let publisherID = tipsList[indexPath.row].id
+    state.ledger.removeRecurringTip(publisherId: publisherID)
+    tipsList.remove(at: indexPath.row)
+    if tipsList.isEmpty {
```
kylehickinson: You don't reload the table if the tipsList is non-empty after removing it? I'm confused as to what happens here

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
```diff
@@ -149,8 +207,15 @@ extension TipsDetailViewController: UITableViewDataSource, UITableViewDelegate {
...
+      tableView.reloadData()
```
kylehickinson: Should probably animate the change rather than abruptly reload the table

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
```diff
@@ -36,6 +39,32 @@ class TipsDetailViewController: UIViewController {
     tipsView.tableView.dataSource = self
 
     title = Strings.Tips
+    loadData()
+  }
+  
+  private func loadData() {
+    _ = getTipsThisMonth().then {
+      totalBatTips = $0.total.isEmpty ? "0.0" : $0.total
```
kylehickinson: You aren't checking for tips here? Total can be zero if they were to tip 10 BAT and then receive 10 BAT from ads/grants.  You should be creating `BATValue` objects from `oneTimeDonation` and `recurringDonation` and checking those

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
danishjafri88: Oh yeah, copied this over from auto contribute :)

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
danishjafri88: I find it unnecessary, but will update :)

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
danishjafri88: Screenshot 2019-06-20 at 10 03 07 PM
Removing them results in this error

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
danishjafri88: Oh I thought total was that. Thanks for the info

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
danishjafri88: Instead of tip or indexpath I updated to using url and id so that we dont strongly capture Tip.

kylehickinson: > Instead of tip or indexpath I updated to using url and id so that we dont strongly capture Tip.
I don't understand what you mean… Just change `[weak self]` to `[weak self, weak tip]` or whatever the syntax is

BraveRewardsUI/Extensions/DateExtensions.swift
danishjafri88: The `reconcileStamp` is `UInt64`, so to use this method we have to cast everytime. And since this method was created for conversion of `reconcileStamp` to string, I kept the param as UInt64

kylehickinson: I realize that but nothing in the name of this specifies it's for reconcile stamp. 

The way it's written (name, location, etc) all just make this a convenience method for using a date formatted from a Unix timestamp.

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
danishjafri88: Ah sorry, forgot to add else condition. Changing to updating the entire table.
If the list is empty we need the empty cell with text and if there is not recurring tip in the list then remove the next contribution cell from 0 section.

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
danishjafri88: Agreed I just took the identifier as thats what we actually need. Saves us the check for weak tip.

BraveRewardsUI/Extensions/DateExtensions.swift
danishjafri88: updated the param to `reconcileStamp`

kylehickinson: Sorry, one more nit: I just realized this is an extension on `Date`, so prefixing with `date` in the method is redundant. Can just be `string(from:)` or `stringFrom(reconcileStamp:)`

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
```diff
@@ -94,15 +124,18 @@ extension TipsDetailViewController: UITableViewDataSource, UITableViewDelegate {
...
     case .tips:
-//      return 1
-      return 3
+      return tipsList.isEmpty ? 1 : tipsList.count
     }
   }
   
   func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
     tableView.deselectRow(at: indexPath, animated: true)
+    guard Section(rawValue: indexPath.section) == .tips,
```
kylehickinson: Crashed here when clicking on the empty state cell, needs a check to see if tipsList is non-empty

kylehickinson: Oh and the empty cell should have `selectionStyle` set to `.none` :)

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
```diff
@@ -36,6 +38,41 @@ class TipsDetailViewController: UIViewController {
...
+  }
+  
+  private var nextContributionDateView: LabelAccessoryView {
+    let view = LabelAccessoryView()
+    let dateFormatter = DateFormatter().then {
+      $0.dateStyle = .short
+      $0.timeStyle = .none
+    }
+    let reconcileDate = Date(timeIntervalSince1970: TimeInterval(state.ledger.autoContributeProps.reconcileStamp))
+    view.label.text = dateFormatter.string(from: reconcileDate)
```
kylehickinson: This is still not using the `Date.stringFrom` API?

danishjafri88: Ah sorry, copied it from you old code to computed property and forgot to change it. Updating

Issue #103, opened 2019-06-20
Support new Pending Contribution API


kylehickinson:
[brave-core/1528](https://github.com/brave/brave-core/pull/1528/files#diff-a134dae29bd859dd8403c83c5b83284bR230) moved pending contribution info into a separate API so they are no longer use `PublisherInfo` to move data around.

We already have empty bridged methods in BATBraveLedger.mm

kylehickinson:
Blocked until https://github.com/brave/brave-core/pull/2757 is merged

Issue #102, opened 2019-06-20
Support new balances API


kylehickinson:
[brave-core/2715](https://github.com/brave/brave-core/pull/2715/files#diff-99374513446ca185ff5e5b99d7d82fd9R296) adds a new data structure called `Balance` that now holds the users balance rather than being part of WalletProperties.

This effectively means we may have to remove our `balance` read-only property in `BATBraveLedger` if this is an async call

Also update `hasSufficientBalanceToReconcile` call to async version

kylehickinson:
Blocked until https://github.com/brave/brave-core/pull/2757 is merged

PR #104, opened 2019-06-24
Update library to include several fixes

kylehickinson:
- Fix possible crash when making a cpp ActivityInfoFilter from Obj-C ver. (brave-core#2690)
- Added LogStream implementation to allow iOS to capture logs from Brav… (brave-core#2719)
- Add callbacks for all database write methods and use callbacks in ledger client methods (brave-core#2767)
- Fix crash if logger callbacks are not set, allow nullable flush callback (brave-core#2776)

**Note that this update breaks:** 
- Balances & rates (BAT ↔︎ USD) (#102)
- Pending contribution related API (#103)

Please do not merge until after TF build is built and uploaded


BraveRewardsExample/ViewController.swift
```diff
@@ -70,6 +70,14 @@ let stateStoragePath: String = (NSSearchPathForDirectoriesInDomains(
     true
   ).first! as NSString).appendingPathComponent("brave_ledger")
 
+func rewardsLog(logLevel: LogLevel, line: Int32, file: String, data: String) {
+  if !data.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty {
+    // Should probably just trim the final trailing newline instead, since `print` only appends
+    // a single newline terminator
+    print("[\(logLevel.logPrefix)] \(data.trimmingCharacters(in: .newlines))")
```
iccub: Do we need to implement it on brave-ios go get control of rewards logs?

kylehickinson: Yes, though the brave-ios side will use XCLogger instead of `logPrefix`

danishjafri88:
Looks good

BraveRewardsExample/ViewController.swift
kylehickinson: re: os_log vs print, for this case print is fine because this is just an example/host for development, and won't be used in production ever

Issue #62, opened 2019-05-21
Add terms of service disclaimer when enabling Brave Rewards


kylehickinson:
- [x] Below the welcome/learn more screen buttons
- [x] In the panel below the enable Brave Rewards button (after user manually disables Brave Rewards)
- [x] Below the create wallet button?
- [x] Add new localized Strings to Strings.swift

Desktop refs:

Disabled rewards:
<img width="378" alt="desktop-ref" src="https://user-images.githubusercontent.com/529104/59072416-76203080-8890-11e9-98a0-cd1f81905843.png">

Create wallet:
<img width="361" alt="desktop-ref2" src="https://user-images.githubusercontent.com/529104/59072418-77515d80-8890-11e9-911e-d9be870b6a07.png">

On Learn more about brave rewards screen:
<img width="868" alt="desktop-ref3" src="https://user-images.githubusercontent.com/529104/59072420-78828a80-8890-11e9-9281-03e28a978696.png">
<img width="795" alt="desktop ref 4" src="https://user-images.githubusercontent.com/529104/59072423-79b3b780-8890-11e9-9b67-2b318cc22fa4.png">

PR #106, opened 2019-06-25
Added disclaimer label to BraveRewardsUI.

Brandon-T:
Resolves: https://github.com/brave/brave-rewards-ios/issues/62

- Rebased
- Added DisclaimerLabel class
- Documented Code
- Added DisclaimerLabel UI to:
  - Below the welcome/learn more screen buttons
  - In the panel below the enable Brave Rewards button (after user manually disables Brave Rewards)
  - Below the create wallet button?
  - Add new localized Strings to Strings.swift


BraveRewardsUI/Common/DisclaimerLinkLabel.swift
```diff
@@ -0,0 +1,134 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+
+/// A Disclaimer Label that allows clickable links (or data-detectors)
+final class DisclaimerLinkLabel: UITextView {
```
kylehickinson: This class is a both generic and specific, for example: We don't set the text in it despite it always being the same for its usages, but we specify `Link` with only choices that are available with that text, we force centre alignment, etc. So I'd say we should pick a direction for it. Either we:

- Make it fully generic, allow the `onLinkedTapped` to just give back the URL, let the controller handle it (or pass in some sort of Generic string enum which maps case to url. Let the call-site choose styling, etc. Then replace `DisclaimerView` with this.
- Make it specific to ToS & PP, giving it specific text and move it out of the Common folder

BraveRewardsUI/Common/DisclaimerLinkLabel.swift
```diff
@@ -0,0 +1,134 @@
...
+    let attributedString = { () -> NSAttributedString? in
+      let paragraphStyle = NSMutableParagraphStyle()
+      paragraphStyle.alignment = .center
+      
+      let attributes: [NSAttributedString.Key: Any] = [.font: self.font ?? UIFont.systemFont(ofSize: 12.0),
+                                                       .foregroundColor: self.textColor ?? UX.textColor,
+                                                       .paragraphStyle: paragraphStyle]
+      
+      let text = try? NSMutableAttributedString(data: text.data(using: .utf8) ?? Data(),
+                                                          options: [.documentType: NSAttributedString.DocumentType.html,
```
kylehickinson: Is this still performant using HTML parsing over just specifying `.link` attributes manually?

BraveRewardsUI/Common/DisclaimerLinkLabel.swift
```diff
@@ -0,0 +1,134 @@
...
+    
+    let attributedString = { () -> NSAttributedString? in
+      let paragraphStyle = NSMutableParagraphStyle()
+      paragraphStyle.alignment = .center
+      
+      let attributes: [NSAttributedString.Key: Any] = [.font: self.font ?? UIFont.systemFont(ofSize: 12.0),
+                                                       .foregroundColor: self.textColor ?? UX.textColor,
+                                                       .paragraphStyle: paragraphStyle]
+      
+      let text = try? NSMutableAttributedString(data: text.data(using: .utf8) ?? Data(),
```
kylehickinson: Probably just bail instead of pass in empty Data

BraveRewardsUI/Common/DisclaimerLinkLabel.swift
```diff
@@ -0,0 +1,134 @@
...
+                                                          options: [.documentType: NSAttributedString.DocumentType.html,
+                                                                    .defaultAttributes: attributes],
+                                                          documentAttributes: nil)
+      
+      let range = NSRange(location: 0, length: text?.length ?? 0)
```
kylehickinson: We should just `guard let text = text else { return nil }` here as you're using `text?` an awful lot here and theres no reason to do anything if it returns nil for some reason

BraveRewardsUI/Common/DisclaimerLinkLabel.swift
```diff
@@ -0,0 +1,134 @@
...
+      text?.addAttribute(.foregroundColor, value: self.textColor ?? UX.textColor, range: range)
+      text?.addAttribute(.paragraphStyle, value: paragraphStyle, range: range)
```
kylehickinson: Wouldn't it make sense to add these 2 lines in the `beginEditing`/`endEditing` block?

BraveRewardsUI/Common/DisclaimerLinkLabel.swift
```diff
@@ -0,0 +1,134 @@
...
+    self.linkTextAttributes = linkAttributes
+    self.attributedText = attributedString()
+    
+    setAccessibility()
+  }
+  
+  /// Makes this label accessible as static text.
+  private func setAccessibility() {
+    accessibilityLabel = self.text
+    accessibilityTraits = .staticText
```
kylehickinson: Is `staticText` correct here? It technically has buttons on it

BraveRewardsUI/Common/DisclaimerLinkLabel.swift
Brandon-T: Made it more generic.

BraveRewardsUI/Common/DisclaimerLinkLabel.swift
Brandon-T: Fixed.

BraveRewardsUI/Common/DisclaimerLinkLabel.swift
Brandon-T: Fixed.

BraveRewardsUI/Common/DisclaimerLinkLabel.swift
Brandon-T: I haven't noticed any performance issues since the text is very small and we're setting it only once on creation of the labels. In a tableview it might be a problem.

BraveRewardsUI/Common/DisclaimerLinkLabel.swift
Brandon-T: I got rid of those lines and just passed in a dictionary and also placed it between the begin and end.

BraveRewardsUI/Common/DisclaimerLinkLabel.swift
Brandon-T: Made it way more generic.

BraveRewardsUI/Create Wallet/CreateWalletView.swift
```diff
@@ -105,17 +113,36 @@ extension CreateWalletViewController {
...
+      
+      termsOfServiceLabel.onLinkedTapped = onLinkTapped
     }
     
     @available(*, unavailable)
     required init(coder: NSCoder) {
       fatalError()
     }
+    
+    private func onLinkTapped(_ link: URL) {
```
kylehickinson: This should actually be handled at a controller level (i.e. `CreateWalletController`), and should call the Rewards delegate method that asks the client to open a URL (`RewardsUIDelegate` → `loadNewTabWithURL(_:)`).

Also, the URLs are: 
- ToS: https://brave.com/terms-of-use/
- PP: https://brave.com/privacy/#rewards

BraveRewardsUI/Wallet/RewardsSummaryProtocol.swift
```diff
@@ -70,13 +70,18 @@ extension RewardsSummaryProtocol {
...
     if reservedAmount.doubleValue <= 0 { return nil }
     
     let text = String(format: Strings.ContributingToUnverifiedSites, reservedAmount.displayString)
     
-    return DisclaimerView(text: text)
+    return LinkLabel().then {
+      $0.textColor = Colors.grey200
+      $0.font = UIFont.systemFont(ofSize: 12.0)
+      $0.textContainerInset = UIEdgeInsets(top: 8.0, left: 8.0, bottom: 8.0, right: 8.0)
+      $0.text = "\(text) \(Strings.DisclaimerLearnMore)"
```
kylehickinson: Same as above regarding background color + corner radius

BraveRewardsUI/Tipping/TippingOverviewView.swift
```diff
@@ -48,9 +48,13 @@ class TippingOverviewView: UIView {
...
   
-  let disclaimerView = DisclaimerView(text: Strings.TippingUnverifiedDisclaimer).then {
-    $0.isHidden = true
+  let disclaimerView = LinkLabel().then {
+    $0.textColor = Colors.grey200
+    $0.font = UIFont.systemFont(ofSize: 12.0)
+    $0.textContainerInset = UIEdgeInsets(top: 8.0, left: 8.0, bottom: 8.0, right: 8.0)
+    $0.text = "\(Strings.TippingUnverifiedDisclaimer) \(Strings.DisclaimerLearnMore)"
     $0.backgroundColor = .white
+    $0.isHidden = true
```
kylehickinson: Missing background color + corner radius as. it was before:

https://github.com/brave/brave-rewards-ios/blob/a9a15960f87b3577c6db036519782d5b4eebbced/BraveRewardsUI/Common/DisclaimerView.swift#L37-L38

If this isn't possible to fix, perhaps we revert the deletion of `DisclaimerView` and simply embed a `LinkLabel` within it?

BraveRewardsUI/Welcome/WelcomeView.swift
```diff
@@ -13,6 +13,13 @@ extension WelcomeViewController {
       $0.contentEdgeInsets = UIEdgeInsets(top: 8, left: 12, bottom: 8, right: 12)
     }
     
+    let termsOfServiceLabel = LinkLabel().then {
+      $0.font = .systemFont(ofSize: 12.0)
+      $0.textColor = Colors.grey900
+      $0.textAlignment = .center
+      $0.text = Strings.DisclaimerInformation
```
kylehickinson: Wrong string here, there's a variant of the original message for the welcome screen:

`By clicking ‘Yes, I'm in!’, you indicate that you have read and agree to the Terms of Service and Privacy Policy.`

kylehickinson:
Also missing the second addition to WelcomeView below the "yes I'm ready" create wallet button at the bottom of the screen as seen in https://github.com/brave/brave-rewards-ios/issues/62

BraveRewardsUI/Create Wallet/CreateWalletView.swift
Brandon-T: Fixed.

Brandon-T:
> Also missing the second addition to WelcomeView below the "yes I'm ready" create wallet button at the bottom of the screen as seen in #62

Fixed :)
I didn't even notice that lol. Thanks!

BraveRewardsUI/Publisher/PublisherView.swift
```diff
@@ -84,7 +86,12 @@ class PublisherView: UIStackView {
     $0.adjustsFontSizeToFitWidth = true
   }
   // Only shown when unverified
-  private let unverifiedDisclaimerView = DisclaimerView(text: Strings.UnverifiedPublisherDisclaimer)
+  private let unverifiedDisclaimerView = LinkLabel().then {
+    $0.textColor = Colors.grey200
+    $0.font = UIFont.systemFont(ofSize: 12.0)
+    $0.textContainerInset = UIEdgeInsets(top: 8.0, left: 8.0, bottom: 8.0, right: 8.0)
+    $0.text = "\(Strings.UnverifiedPublisherDisclaimer) \(Strings.DisclaimerLearnMore)"
```
kylehickinson: Missing background color + radius here, sorry missed this one, its the same as the other replaced DisclaimerView (`UIColor(white: 0.0, alpha: 0.04)`)

Simulator Screen Shot - iPhone 8 - 2019-06-25 at 16 40 08

BraveRewardsUI/Welcome/WelcomeViewController.swift
```diff
@@ -67,4 +71,17 @@ class WelcomeViewController: UIViewController {
       })
     }
   }
+  
+  private func tappedDisclaimerLink(_ url: URL) {
+    switch url.path {
+    case "/terms":
+      state.delegate?.loadNewTabWithURL(URL(string: "https://brave.com/terms-of-use/")!) //swiftlint:disable:this force_unwrapping
```
kylehickinson: We may want to put make these URLs static and use them in multiple places in case the URL ever changes

BraveRewardsUI/Create Wallet/CreateWalletViewController.swift
```diff
@@ -79,4 +80,17 @@ class CreateWalletViewController: UIViewController {
     let controller = WelcomeViewController(state: state)
     navigationController?.pushViewController(controller, animated: true)
   }
+  
+  private func tappedDisclaimerLink(_ url: URL) {
+    switch url.path {
+    case "/terms":
+      state.delegate?.loadNewTabWithURL(URL(string: "https://brave.com/terms-of-use/")!) //swiftlint:disable:this force_unwrapping
```
iccub: Doesn't hurt to unwrap it. It's better to do nothing than to crash the app

BraveRewardsUI/Create Wallet/CreateWalletViewController.swift
```diff
@@ -79,4 +80,17 @@ class CreateWalletViewController: UIViewController {
...
+      
+    case "/policy":
+      state.delegate?.loadNewTabWithURL(URL(string: "https://brave.com/privacy/#rewards")!) //swiftlint:disable:this force_unwrapping
```
iccub: Same as above, unwrap

BraveRewardsUI/Create Wallet/CreateWalletViewController.swift
```diff
@@ -79,4 +80,17 @@ class CreateWalletViewController: UIViewController {
...
+      
+    default:
+      break
```
iccub: Add an assertionFailure here

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -295,4 +297,17 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
       rewardsSummaryView.alpha = 1.0
     }
   }
+  
+  private func tappedDisclaimerLink(_ url: URL) {
+    switch url.path {
+    case "/terms":
+      state.delegate?.loadNewTabWithURL(URL(string: "https://brave.com/terms-of-use/")!) //swiftlint:disable:this force_unwrapping
```
iccub: Unwrap, and extract the url string somewhere? We use it in two places, would be good to put it somewhere so when it changes it will require only one place to fix

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -295,4 +297,17 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
...
+      
+    case "/policy":
+      state.delegate?.loadNewTabWithURL(URL(string: "https://brave.com/privacy/#rewards")!) //swiftlint:disable:this force_unwrapping
```
iccub: Unwrap and extract url string somewhere

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -295,4 +297,17 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
...
+      
+    default:
+      break
```
iccub: Add assertionFailure

BraveRewardsUI/Welcome/WelcomeViewController.swift
```diff
@@ -67,4 +71,17 @@ class WelcomeViewController: UIViewController {
       })
     }
   }
+  
+  private func tappedDisclaimerLink(_ url: URL) {
```
iccub: this `tappedDisclaimerLink` method looks the same in 3 view controllers, perhaps we could share it in one place?
Idk, maybe a protocol for that would be good, maybe it's an overkill

Issue #65, opened 2019-05-21
Add ability to force update publisher list


kylehickinson:
Desktop has "Check again…" button next to unverified publishers which forces the ledger to re-download the latest publisher list

Desktop design:
<img width="376" alt="desktop-ref" src="https://user-images.githubusercontent.com/529104/59072260-f2feda80-888f-11e9-9163-68b2bc162ffd.png">

PR #107, opened 2019-06-25
Added CheckAgain button to reload publisher verification status.

Brandon-T:
Resolves: https://github.com/brave/brave-rewards-ios/issues/65

- Rebased
- Added Check Again Button
- Refresh Publisher Verification Status on "Check Again" tapped.

### DIFF
--- a/BraveRewardsUI/Publisher/PublisherView.swift
+++ b/BraveRewardsUI/Publisher/PublisherView.swift
@@ -49,6 +49,9 @@ class PublisherView: UIStackView {
     }
   }
   
+  /// Refresh Publisher List
+  var onCheckAgainTapped: (() -> Void)?
+  
   // MARK: -
   
   private struct UX {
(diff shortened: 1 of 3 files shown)


BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -156,17 +157,25 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
...
-        guard let publisher = info else { return }
-        assert(Thread.isMainThread)
-        publisherView.setVerified(publisher.verified)
+      publisherView.onCheckAgainTapped = { [weak self] in
+        guard let self = self else { return }
         
-        if let percent = self.state.ledger.currentActivityInfo(withPublisherId: publisher.id)?.percent {
-          attentionView.valueLabel.text = "\(percent)%"
+        publisherView.checkAgainButton.isLoading = true
+        self.state.ledger.publisherInfo(forId: host) { info in
```
kylehickinson: Hah, this is actually not the call that has to be made here. Its an API called `RefreshPublisher` which we haven't wrapped yet in `BATBraveLedger`, so I'll get you to update this when that API is available

kylehickinson:
Blocked until we can update the library

BraveRewardsUI/Publisher/PublisherView.swift
```diff
@@ -69,20 +72,38 @@ class PublisherView: UIStackView {
...
+  }
+  
   // "Brave Verified Publisher" / "Not yet verified"
   private let verifiedLabel = UILabel().then {
     $0.textColor = UX.verifiedStatusColor
     $0.font = .systemFont(ofSize: 12.0)
     $0.adjustsFontSizeToFitWidth = true
   }
+  
+  // "Brave Check Again" button
```
iccub: This comment can be removed, it doesn't say anything useful, the button name below is good enoguh

BraveRewardsUI/Publisher/PublisherView.swift
```diff
@@ -69,20 +72,38 @@ class PublisherView: UIStackView {
...
     $0.spacing = 4.0
   }
+  
   // ✓ or ?
   private let verificationSymbolImageView = UIImageView().then {
+    $0.contentMode = .scaleAspectFit
     $0.setContentHuggingPriority(.required, for: .horizontal)
   }
+  
+  // For containing verifiedLabel and checkAgainButton
```
iccub: This comment also doesn't add anything useful imo

Brandon-T:
> Just tried this out on Desktop, and noticed that when you click on it, it ~shows~ replaces the button with a loader and after completing, it actually hides the button entirely. For the loader we can just use the custom `LoaderView` already built

Weird.. I'm using the `LoaderView(size: .small)` on the `Button` class.

kylehickinson:
> > Just tried this out on Desktop, and noticed that when you click on it, it ~shows~ replaces the button with a loader and after completing, it actually hides the button entirely. For the loader we can just use the custom `LoaderView` already built
> 
> Weird.. I'm using the `LoaderView(size: .small)` of the `Button` class.

Ah! Then you're good, sorry I didn't notice that. Yeah should work then, just have to make it fully hide the button when its done then

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -156,17 +157,29 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
...
+          publisherView.checkAgainButton.isLoading = false
+          
+          guard let publisher = info else { return }
+          assert(Thread.isMainThread)
+          publisherView.setVerified(publisher.verified)
+          
+          if publisher.verified {
+            publisherView.checkAgainButton.isHidden = true
```
kylehickinson: This should hide regardless of if the refreshed publisher is verified

BraveRewardsUI/Wallet/WalletViewController.swift

BraveRewardsUI/Wallet/WalletViewController.swift

PR #109, opened 2019-06-26
Fix #97 Hiding publisher view by showing rewardsSummaryView on top when url is local

### LINKED ISSUE #97: Localhost url should hide publisher info
With url provided: `"http://localhost:6571/about/home/#panel=0"`

actual:

Zrzut ekranu 2019-06-17 o 23 50 31

expected(summary only)
Zrzut ekranu 2019-06-18 o 00 01 27

danishjafri88:
Fix #97 Hiding publisher view by showing rewardsSummaryView on top when url is local.

### DIFF
--- a/BraveRewardsUI/Publisher/PublisherSummaryView.swift
+++ b/BraveRewardsUI/Publisher/PublisherSummaryView.swift
@@ -80,6 +80,10 @@ class PublisherSummaryView: UIView {
 }
 
 extension PublisherSummaryView: WalletContentView {
+  var estimatedHeight: CGFloat {
+    return scrollView.contentSize.height
+  }
+  
   var innerScrollView: UIScrollView? {
     return scrollView
   }
--- a/BraveRewardsUI/Publisher/PublisherSummaryView.swift
+++ b/BraveRewardsUI/Publisher/PublisherSummaryView.swift
@@ -10,15 +10,6 @@ class PublisherSummaryView: UIView {
     $0.textLabel.text = Strings.TipSiteMonthly
   }
   
-  func setLocal(_ local: Bool) {
-    if local {
-      publisherView.publisherNameLabel.text = "Brave Browser"
-      publisherView.faviconImageView.image = UIImage(frameworkResourceNamed: "local-icon")
-      attentionView.valueLabel.text = "–"
-    }
-    publisherView.faviconImageView.contentMode = local ? .center : .scaleAspectFill
-  }
-  
   private let scrollView = UIScrollView().then {
     $0.contentInsetAdjustmentBehavior = .never
     $0.delaysContentTouches = false
@@ -51,7 +42,16 @@ class PublisherSummaryView: UIView {
   
   public override init(frame: CGRect) {
     super.init(frame: frame)
-    
+  }
+  
+  convenience init(emptyView: Bool) {
+    self.init()
+    if !emptyView {
+      setup()
+    }
+  }
+  
+  private func setup() {
     addSubview(scrollView)
     scrollView.addSubview(stackView)
     
--- a/BraveRewardsUI/Publisher/PublisherSummaryView.swift
+++ b/BraveRewardsUI/Publisher/PublisherSummaryView.swift
@@ -42,16 +42,6 @@ class PublisherSummaryView: UIView {
   
   public override init(frame: CGRect) {
     super.init(frame: frame)
-  }
-  
-  convenience init(emptyView: Bool) {
-    self.init()
-    if !emptyView {
-      setup()
-    }
-  }
-  
-  private func setup() {
     addSubview(scrollView)
     scrollView.addSubview(stackView)
     
--- a/BraveRewardsUI/Wallet/WalletView.swift
+++ b/BraveRewardsUI/Wallet/WalletView.swift
@@ -141,10 +141,26 @@ extension WalletViewController {
       
       summaryLayoutGuide.snp.makeConstraints { $0.bottom.equalTo(self) }
     }
+    
+    func estimatedHeight(hideConetent: Bool) -> CGFloat {
+      var height = headerView.bounds.height
+      if hideConetent {
+        height += rewardsSummaryView.estimatedHeight +
+        20// Padding between Header and content/rewardsSummary
+      } else if contentView?.displaysRewardsSummaryButton == true {
+        height += rewardsSummaryView.rewardsSummaryButton.bounds.height +
+          (contentView?.estimatedHeight ?? 0.0)
+      } else {
+        height += (contentView?.estimatedHeight ?? 0.0)
+      }
+      return height
+    }
   }
 }
 
-class EmptyWalletContentView: UIView, WalletContentView {
+class EmptyContentView: UIView, WalletContentView {
+  var estimatedHeight: CGFloat { return 0 }
+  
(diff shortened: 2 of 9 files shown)


danishjafri88:
> Found a bug when the wallet is newly created: [bug.mov.zip](https://github.com/brave/brave-rewards-ios/files/3330218/bug.mov.zip)
> 
> Also this looks… not great… when its empty:
> Simulator Screen Shot - iPhone 8 - 2019-06-26 at 09 31 18
> 
> Maybe we can force the overall size to be lower for the panel if the balance report is empty and we're on a local host?

Sorry for the delay. Was trying to check as many cases as possible. The layout change is tricky due to the design and subsequent constraints. Last commit makes sure they work well. Please test and lmk if I missed something.

BraveRewardsUI/Publisher/PublisherSummaryView.swift
```diff
@@ -51,7 +42,16 @@ class PublisherSummaryView: UIView {
   
   public override init(frame: CGRect) {
     super.init(frame: frame)
-    
+  }
+  
+  convenience init(emptyView: Bool) {
+    self.init()
+    if !emptyView {
```
kylehickinson: So the fact that we do this just makes me think this logic should be split out and set as a whole new view in `WalletViewController` same as we do here:

https://github.com/brave/brave-rewards-ios/blob/ef5ac14d32230dea09f6b6aa5122ae875a6dd8a1/BraveRewardsUI/Wallet/WalletViewController.swift#L205-L216

Basically just adding a check for local within the `isEnabled` flow and setting `contentView` to a `RewardsSummaryView` 

This would let us take away all these odd changes for handling layout and such in the publisher view stuff since its separate anyways and clean things up significantly here. Should make layout logic easier too

danishjafri88: Sounds good! Will add

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -111,27 +111,47 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
...
+      }
     }
     
     let newSize = CGSize(width: RewardsUX.preferredPanelSize.width, height: height)
     if preferredContentSize != newSize {
       preferredContentSize = newSize
     }
+    if isLocal {
+      transformRewardSummaryView(isExpanding: true, animated: false)
+      walletView.rewardsSummaryView.rewardsSummaryButton.slideToggleImageView.image = nil
```
kylehickinson: You should override `isDisabled` on `RewardsSummaryButton` and _hide_ the image (instead of setting it to nil) based on the state

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -111,27 +111,47 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
...
-      
-      height += scrollView.contentSize.height
-    } else {
-      height += contentView.bounds.height
-    }
-    
+    var height: CGFloat
     if state.ledger.isEnabled {
-      height = RewardsUX.preferredPanelSize.height
+      if summaryRows.isEmpty && isLocal {
```
kylehickinson: All of this additional layout logic seems incorrect, if you add a size to the empty content view you created is this needed?

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -257,19 +275,25 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
...
   
   @objc private func tappedRewardsSummaryButton() {
-    let contentView = walletView.contentView
     let rewardsSummaryView = walletView.rewardsSummaryView
     
     let isExpanding = rewardsSummaryView.transform.ty == 0
+    transformRewardSummaryView(isExpanding: isExpanding, animated: true)
+  }
+  
+  private func transformRewardSummaryView(isExpanding: Bool, animated: Bool) {
```
kylehickinson: intent/apple naming: `transformRewardSummaryView` → `setRewardsSummaryVisible(_ visible: Bool, animated: Bool = false)`

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -257,19 +275,25 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
...
       UIImage(frameworkResourceNamed: isExpanding ? "slide-down" : "slide-up")
     
     // Animating the rewards summary with a bit of a bounce
-    UIView.animate(withDuration: 0.4, delay: 0, usingSpringWithDamping: 0.85, initialSpringVelocity: 0, options: [], animations: {
+    UIView.animate(withDuration: animated ? 0.4 : 0.0, delay: 0, usingSpringWithDamping: 0.85, initialSpringVelocity: 0, options: [], animations: {
       if isExpanding {
         rewardsSummaryView.transform = CGAffineTransform(
           translationX: 0,
-          y: -self.walletView.summaryLayoutGuide.layoutFrame.height + rewardsSummaryView.rewardsSummaryButton.bounds.height
+          y: -self.walletView.bounds.height + self.walletView.headerView.bounds.height + 20 + rewardsSummaryView.rewardsSummaryButton.bounds.height
```
kylehickinson: Is there a reason you're no longer using the summary layout guide?

BraveRewardsUI/Wallet/WalletView.swift
```diff
@@ -143,3 +143,13 @@ extension WalletViewController {
     }
   }
 }
+
+class EmptyWalletContentView: UIView, WalletContentView {
```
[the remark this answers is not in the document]
kylehickinson: Also should be `private`

BraveRewardsUI/Wallet/WalletViewController.swift
danishjafri88: Updated. Kept that name as in either case it is still visible, only a change in height.

BraveRewardsUI/Wallet/WalletView.swift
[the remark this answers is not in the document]
danishjafri88: Can't make it private as it wont be accessible from WalletVewController.
We can give it a size but the height of the Wallet is not affected by it. The way to change the height of the WalletView is by changing the preferredContentSize

BraveRewardsUI/Wallet/WalletViewController.swift
danishjafri88: Summary layoutGuide is actually dependent on the preferredContentSize to do the resize. We need a value here before/ during the animation is happening.

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -96,6 +98,11 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
       navigationController?.setNavigationBarHidden(true, animated: animated)
     }
     reloadUIState()
+    
+    // Not sure why this works
+    DispatchQueue.main.asyncAfter(deadline: .now()) {
+      self.configureSize()
```
danishjafri88: Doing this without `asyncAfter` shows no change. Maybe because of the animations in PopoverController's `preferredContentSizeDidChange`

kylehickinson:
Will be done in #113

Issue #47, opened 2019-05-15
Integrate Auto-Contribute Details & List


kylehickinson:
When Auto-Contribute is enabled in settings, you can tap the `View Details` button. On this new screen, we should:

- [x] The next contribution date should be computed correctly
- [x] Populate the list using `listActivityInfoFromStart:limit:filter:completion`, **only showing non-zero attention percentages**)
- [x] The supported sites number should be reported correctly based on the listing
- [x] Media publishers may show a name + `on YouTube`/`on Twitch` instead of domain
- [x] Show the correct attention per client (`BATPublisherInfo.percent`)
- [x] Probably implement some sort of automatic pagination when scrolling to the end of the list
- [ ] Swipe-to-exclude (or tapping edit button then excluding a publisher) should update said publisher with (`updatePublisherExclusionState:state:`)
- [x] If the `numberOfExcludedPublishers` is > 0, we should show the "Restore %d Publishers" cell in the list. Tapping said list will bring up an action sheet. Tapping the confirmation should call `restoreAllExcludedPublishers` ~and then reload the list using `listActivityInfoFromStart:limit:filter:completion` again to possibly show restored publishers. The restore button should disappear now that theres 0 excluded publishers~ Restored publishers start with 0% therefore won't be shown
- [x] Tapping the item currently does nothing but may need to open that URL?

![Simulator Screen Shot - iPhone X - 2019-05-15 at 16 51 09](https://user-images.githubusercontent.com/529104/57808594-aeb96800-7731-11e9-971d-82de8e25f3e4.png)

Desktop ref: https://brave.github.io/brave-ui/?path=/story/feature-components-rewards-table--contribution

PR #114, opened 2019-07-03
Auto-contribute improvements & publisher exclusion

iccub:
Ref #47 

- Autocontribute setting per publisher is implemented, toggle the switch to include or exclude publisher from AC, this changes are reflected on AC settings screen as well.

- Auto contribute settings how show a favicons in the same way as on publishers summary panel

Zrzut ekranu 2019-07-1 o 19 13 23

### DIFF
--- a/BraveRewardsUI/Wallet/WalletViewController.swift
+++ b/BraveRewardsUI/Wallet/WalletViewController.swift
@@ -4,6 +4,7 @@
 
 import UIKit
 import SnapKit
+import BraveRewards
 
 protocol WalletContentView: AnyObject {
   var innerScrollView: UIScrollView? { get }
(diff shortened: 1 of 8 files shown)


BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -201,6 +205,20 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
...
+    publisherSummaryView.autoContributeChanged = { [weak self] enabled in
+      guard let self = self, let host = self.state.url.host else { return }
+      
+      // setting publisher exclusion to .included didn't seem to work, using .default instead.
+      let state: PublisherExclude = enabled ? .default : .excluded
+      self.state.ledger.updatePublisherExclusionState(withId: host, state: state)
+      
+      if !enabled {
+        // Disabled auto-contribute resets attention value.
+        attentionView.valueLabel.text = "0%"
```
kylehickinson: This is _probably_ fine to imply, but we might want to instead attempt to retrieve the activity (even though it'll just create a new empty one I think) and set the value based on that.

Issue #46, opened 2019-05-15
Integrate tipping a publisher directly


kylehickinson:
The "Send Tip" button on a publisher brings up the tipping screen. 

- [ ] The header content for this screen are populated by getting the publishers banner (`publisherBannerForId:completion:`) including what social media icons show up, the tipping amounts, the top background image, title, body, etc.
- [x] The balance in the option selector should respect `balance`
- [x] The selected option should show the "Not enough funds" UI if the users balance is lower than the selected option
- [ ] If the user selects "Make this monthly", it will be added as a recurring donation (`addRecurringTipToPublisherWithId:amount:`)
- [ ] Hitting send tip should call `tipPublisherDirectly:amount:currency`
- [ ] The shown "thank you" message should use the publishers favicon
- [ ] If a user has a recurring tip for this publisher already, the monthly tip is removed

![Simulator Screen Shot - iPhone X - 2019-05-15 at 16 48 59](https://user-images.githubusercontent.com/529104/57808439-5d10dd80-7731-11e9-9e59-c1bdf3546567.png)

- [ ] Tip completion screen has different screen depending on if its a monthly tip or one time tip

<img width="2198" alt="Screen Shot 2019-05-21 at 5 12 08 PM" src="https://user-images.githubusercontent.com/529104/58108315-9fb33980-7bb9-11e9-8374-ea6ca268ed80.png">

PR #118, opened 2019-07-03
Feature/tipping publisher

Brandon-T:
Fixes https://github.com/brave/brave-rewards-ios/issues/46

- Fixed crash in `BatBravePublisherBanner` where we used "description" instead of "desc".. "description" in Objective-C cannot be set and if you want a field named "description", we'd have to explicitly synthesize it and add a setting/getter. I decided to rename the field instead.

- Adds integration for Publisher API with TippingViewController.


BraveRewardsUI/Tipping/TippingViewController.swift
```diff
@@ -63,6 +72,73 @@ class TippingViewController: UIViewController, UIViewControllerTransitioningDele
...
+                                                              .twitter: "pub-twitter",
+                                                              .twitch: "pub-twitch"]
+            
+            if let icon = icons[mediaType] {
+              self.tippingView.overviewView.socialStackView.addArrangedSubview(UIImageView(image: UIImage(frameworkResourceNamed: icon)))
+            }
+          }
+        })
+      } else {
+        ["pub-youtube", "pub-twitter", "pub-twitch"].forEach({
```
kylehickinson: If its empty I don't think we're supposed to show any of these

BraveRewardsUI/Tipping/TippingViewController.swift
```diff
@@ -72,9 +148,31 @@ class TippingViewController: UIViewController, UIViewControllerTransitioningDele
...
-    DispatchQueue.main.asyncAfter(deadline: .now() + 2.0) { [weak self] in
-      self?.dismiss(animated: true)
+    if let selectedIndex = self.tippingView.optionSelectionView.selectedOptionIndex {
+      let amount = self.tippingView.optionSelectionView.options[selectedIndex].value.doubleValue
+      
+      state.ledger.listRecurringTips { [weak self] info in
+        guard let self = self else { return }
+        
+        // If the user has recurring tips, remove the monthly tip
+        if info.firstIndex(where: { $0.id == self.publisherInfo.id }) != nil {
```
kylehickinson: Is this logic taken from brave-core?

BraveRewardsUI/Tipping/TippingViewController.swift
```diff
@@ -72,9 +148,31 @@ class TippingViewController: UIViewController, UIViewControllerTransitioningDele
...
+        }
+        
+        // Add recurring tips if there is none..
+        if self.tippingView.optionSelectionView.isMonthly {
+          self.state.ledger.addRecurringTip(publisherId: self.publisherInfo.id, amount: amount)
+        }
+        
+        // TODO: Figure out -- Not sure why amount is of type integer..
+        // Also not sure user's currency
+        self.state.ledger.tipPublisherDirectly(self.publisherInfo, amount: Int32(amount), currency: "USD")
```
kylehickinson: Currency is actually supposed to be "BAT". We should probably make this a default parameter on `BATBraveLedger` side

BraveRewardsUI/Tipping/TippingViewController.swift
```diff
@@ -72,9 +148,31 @@ class TippingViewController: UIViewController, UIViewControllerTransitioningDele
...
+        self.tippingView.setTippingConfirmationVisible(true, animated: true)
```
kylehickinson: The tipping confirmations screen has the correct favicon now? And I believe it can have different messages appear based on whether or not this was a monthly tip or not

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -279,7 +286,11 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
   }
   
   @objc private func tappedSendTip() {
-    let controller = TippingViewController(state: state, publisherId: "") // TODO: pass publisher Id
+    // TODO: Figure out state when there's no publisher
+    // Not sure what to do if there's no publisher info..
+    // After all, you need their ID to send a tip..
+    guard let publisherInfo = self.publisherInfo else { return }
+    let controller = TippingViewController(state: state, publisherInfo: publisherInfo)
```
kylehickinson: Honestly I'd say its probably better to just re-query the publisher info via `state.ledger.publisherInfo`, as when a publishers exclusion state changes it won't be changed in the local variable

BraveRewardsUI/Tipping/TippingViewController.swift
```diff
@@ -5,14 +5,20 @@
 import UIKit
 import BraveRewards
 
+enum BravePublisherSocialMedia: String {
```
iccub: We should use same naming as desktop for that which seems to be `MediaType` or `PublisherMediaType`, cc @kylehickinson for awareness

On desktop I see stuff like `#define YOUTUBE_MEDIA_TYPE "youtube"` in `static_values.h`

BraveRewardsUI/Tipping/TippingViewController.swift
```diff
@@ -63,6 +72,73 @@ class TippingViewController: UIViewController, UIViewControllerTransitioningDele
...
+      })
+      
+      self.tippingView.overviewView.socialStackView.arrangedSubviews.forEach({ $0.removeFromSuperview() })
+      
+      if !banner.social.isEmpty {
+        banner.social.keys.forEach({
+          //When testing with "3zsistemi.si", the values come back empty, but we still have valid keys..
+          //It seems we don't care about the values as long as the key exists, that user has this media..
+          if let mediaType = BravePublisherSocialMedia(rawValue: $0) {
+            let icons: [BravePublisherSocialMedia: String] = [.youtube: "pub-youtube",
```
iccub: Instead of this dictionary we could create a helper in `BravePublisherSocialMedia` enum that will return correct icon, something like

````
// in BravePublisherSocialMedia:
var iconName: String {
  switch self {
    case .youtube: return "pub-youtube"
    ...
  }
}
```

and then use it at line 107 as `icon.iconName`

BraveRewardsUI/Tipping/TippingViewController.swift
```diff
@@ -63,6 +72,73 @@ class TippingViewController: UIViewController, UIViewControllerTransitioningDele
...
+          }
+        })
+      } else {
+        ["pub-youtube", "pub-twitter", "pub-twitch"].forEach({
+          self.tippingView.overviewView.socialStackView.addArrangedSubview(UIImageView(image: UIImage(frameworkResourceNamed: $0)))
+        })
+      }
+      
+      self.tippingView.overviewView.disclaimerView.isHidden = banner.verified
+      self.tippingView.optionSelectionView.options = (banner.amounts.isEmpty ? [1, 5, 10] : banner.amounts).map {
```
iccub: Extract `[1, 5, 10]` to a constant somewhere at file level, it is a bit hidden here

BraveRewardsUI/Tipping/TippingViewController.swift
```diff
@@ -63,6 +72,73 @@ class TippingViewController: UIViewController, UIViewControllerTransitioningDele
...
+      guard let self = self, let banner = banner else { return }
+      
+      self.tippingView.overviewView.titleLabel.text = banner.title.isEmpty ? Strings.TippingOverviewTitle : banner.title
+      self.tippingView.overviewView.bodyLabel.text = banner.desc.isEmpty ? Strings.TippingOverviewBody : banner.desc
+      
+      self.downloadImage(url: banner.background, { image in
+        self.tippingView.overviewView.headerView.image = image
+      })
+      
+      self.downloadImage(url: self.publisherInfo.faviconUrl, { image in
```
iccub: Does this work for you, does it download correct favicon?

In other places we use a delegate method to retrieve favicon `state.dataSource?.retrieveFavicon(with: faviconURL, completion: {[weak self] favData in`
cc @kylehickinson for visibility

BraveRewardsUI/Tipping/TippingViewController.swift
Brandon-T: Fixed.

BraveRewardsUI/Tipping/TippingViewController.swift
Brandon-T: Fixed.

BraveRewardsUI/Tipping/TippingViewController.swift
Brandon-T: Fixed.

BraveRewardsUI/Wallet/WalletViewController.swift

BraveRewardsUI/Tipping/TippingViewController.swift
Brandon-T: Uh I wrote it. I didn't take anything from Core. I took the information from the ticket and made the calls. To figure out which publishers have recurring tips, I made the only call that I saw had "recurringTips" and compared with our publisher to see if our publisher had tips. I couldn't find any fields for that.

BraveRewardsUI/Tipping/TippingViewController.swift
Brandon-T: Fixed.

BraveRewardsUI/Tipping/TippingViewController.swift
Brandon-T: Fixed.

BraveRewardsUI/Tipping/TippingViewController.swift
Brandon-T: Fixed.

Brandon-T:
For the favIconURL from publisher.. it sends either blank or "clear" depending on the publisher.. However, I noticed publishers have a "logo" field.. for example: "Picture of Steve Jobs" as the icon. I would think that's correct to use but not sure.

BraveRewardsUI/Tipping/TippingView.swift
```diff
@@ -11,6 +12,40 @@ extension TippingViewController {
...
+          paragraphStyle.lineBreakMode = .byWordWrapping
+          paragraphStyle.lineSpacing = 8.0
+          
+          let text = NSMutableAttributedString(string: "\(Strings.TippingRecurringDetails)\n", attributes: [
+            .font: UIFont.systemFont(ofSize: 12.0, weight: .regular),
+            .foregroundColor: Colors.grey600
+            ])
+          
+          // TODO: Find out where to get the date!
+          text.append(NSAttributedString(string: "6/20/2019", attributes: [
```
iccub: There's `reconcileStamp` in PublisherInfo that should be used, but I am not sure about logic of it.
I guess after adding a recurring tip, you have to retrieve a publisher again and read its `reconcileStamp`

summoning @kylehickinson for help on this one

BraveRewardsUI/Tipping/TippingViewController.swift
```diff
@@ -41,11 +61,16 @@ class TippingViewController: UIViewController, UIViewControllerTransitioningDele
...
+    tippingView.optionSelectionView.optionChanged = { [weak self] option in
+      guard let self = self else { return }
       if let balanceTotal = self.state.ledger.balance?.total {
         let hasEnoughBalanceForTip = option.value.doubleValue < balanceTotal
         self.tippingView.optionSelectionView.setEnoughFundsAvailable(hasEnoughBalanceForTip)
+      } else {
+        self.tippingView.optionSelectionView.setEnoughFundsAvailable(false)
       }
+      
+      self.tippingView.optionSelectionView.setEnoughFundsAvailable(true)
```
iccub: This is only for testing right? so we can check out tipping flow without having BAT

BraveRewardsUI/Tipping/TippingViewController.swift
```diff
@@ -60,9 +85,69 @@ class TippingViewController: UIViewController, UIViewControllerTransitioningDele
     
     tippingView.optionSelectionView.setWalletBalance(state.ledger.balanceString, crypto: "BAT")
     // FIXME: Remove fake data
-    tippingView.optionSelectionView.options = [1.0, 5.0, 10.0].map {
+    tippingView.optionSelectionView.options = TippingViewController.defaultDollarAmounts.map {
```
iccub: These are BAT amounts actually

BraveRewardsUI/Tipping/TippingView.swift
Brandon-T: Hmm I tried this, it always returns 0 for me. So when I calculate the date since 1970, it'll return 1970 lol.

danishjafri88: @Brandon-T the right time is: `Date(timeIntervalSince1970: TimeInterval(state.ledger.autoContributeProps.reconcileStamp))`

BraveRewardsUI/Tipping/TippingViewController.swift
```diff
@@ -72,9 +155,33 @@ class TippingViewController: UIViewController, UIViewControllerTransitioningDele
...
+        // Add recurring tips if there is none..
+        if self.tippingView.optionSelectionView.isMonthly {
+          self.state.ledger.addRecurringTip(publisherId: self.publisherInfo.id, amount: amount)
+        }
+        
+        // TODO: Figure out -- Not sure why amount is of type integer..
+        // Also not sure user's currency
+        self.state.ledger.tipPublisherDirectly(self.publisherInfo, amount: Int32(amount), currency: "BAT")
+        
+        self.tippingView.setInfo(name: self.publisherInfo.name, tipAmount: amount)
```
danishjafri88: This name property does not include suffix string like "on Youtube" etc. to get it please use:
`let provider = " \(publisher.provider.isEmpty ? "" : String(format: Strings.OnProviderText, publisher.provider))"`

and concat this to name.

BraveRewardsUI/Tipping/TippingViewController.swift
```diff
@@ -72,9 +155,33 @@ class TippingViewController: UIViewController, UIViewControllerTransitioningDele
...
+        }
+        
+        // Add recurring tips if there is none..
+        if self.tippingView.optionSelectionView.isMonthly {
+          self.state.ledger.addRecurringTip(publisherId: self.publisherInfo.id, amount: amount)
+        }
+        
+        // TODO: Figure out -- Not sure why amount is of type integer..
+        // Also not sure user's currency
+        self.state.ledger.tipPublisherDirectly(self.publisherInfo, amount: Int32(amount), currency: "BAT")
```
danishjafri88: I guess we can only tip BAT and not a fiat currency. cc @kylehickinson

kylehickinson: Yeah it is also hard-coded in brave-core to `BAT` here

BraveRewardsUI/Tipping/TippingView.swift
kylehickinson: Yeah issue was mock ledger not making a real wallet, so no reconcile stamp

BraveRewardsUI/Tipping/TippingView.swift
```diff
@@ -11,6 +12,39 @@ extension TippingViewController {
     
     var gesturalDismissExecuted: (() -> Void)?
     
+    func setInfo(name: String, tipAmount: Double, recurringDate: String?) {
```
kylehickinson: Lets rename this to something that conveys its for the confirmation view, and lets also make it `private` since its not used anywhere outside of this scope.

Maybe like `updateConfirmationInfo` or something

BraveRewardsUI/Tipping/TippingViewController.swift
kylehickinson: Im just not sure why you remove recurring tips here only just to re-add it?

BraveRewardsUI/Tipping/TippingViewController.swift
```diff
@@ -72,9 +155,46 @@ class TippingViewController: UIViewController, UIViewControllerTransitioningDele
...
+        guard let self = self else { return }
+        
+        // If the user has recurring tips, remove the monthly tip
+        if info.firstIndex(where: { $0.id == self.publisherInfo.id }) != nil {
+          self.state.ledger.removeRecurringTip(publisherId: self.publisherInfo.id)
+        }
+        
+        // Add recurring tips if there is none..
+        if self.tippingView.optionSelectionView.isMonthly {
+          self.state.ledger.addRecurringTip(publisherId: self.publisherInfo.id, amount: amount)
```
kylehickinson: You shouldn't be tipping directly if its a recurring tip, needs a `return` after this line.

Logic from core:

https://github.com/brave/brave-core/blob/150052b89ab409cb0dd933237c9c1395bfc262c7/components/brave_rewards/browser/rewards_service_impl.cc#L2124-L2145

BraveRewardsUI/Tipping/TippingViewController.swift
```diff
@@ -72,9 +155,46 @@ class TippingViewController: UIViewController, UIViewControllerTransitioningDele
...
+          
+          DispatchQueue.main.asyncAfter(deadline: .now() + 2.0) {
+            self.dismiss(animated: true)
+          }
+        }
+        
+        if self.tippingView.optionSelectionView.isMonthly {
+          let date = Date(timeIntervalSince1970: TimeInterval(self.state.ledger.autoContributeProps.reconcileStamp))
+          let dateString = DateFormatter().then {
+            $0.dateFormat = "MM/dd/yyyy"
```
iccub: Do we have to specify date format here? Would be better to have it locale dependent

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -171,9 +171,13 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
       guard let host = state.url.host else { return }
       attentionView.valueLabel.text = "0%"
       
-      state.ledger.publisherInfo(forId: host) { info in
-        guard let publisher = info else { return }
+      state.ledger.publisherInfo(forId: host) { [weak self] info in
+        guard let self = self else { return }
+        
         assert(Thread.isMainThread)
+        publisherView.checkAgainButton.isHidden = info != nil
```
iccub: This will also hide `check again` button if publisher is in database but not verified, right?
we can use `isHidden == publisher.isVerified`(after line 180)

Issue #42, opened 2019-05-15
Show correct publisher information on BraveRewards panel


kylehickinson:
Integrate with ledger to show the correct publisher information for URL provided when making the panel.

This includes the many state of publishers:
- [x] Populating the correct favicon
- [x] When its a media domain it may show `on YouTube`/`on Twitch` after the persons name instead of the domain
- [x] Verified vs unverified
- [x] Unverified publishers show a disclaimer, handling the "learn more" on it
- [x] Auto-Contribute enabled or not (should correctly update said publisher's contribution type)
- [x] Showing if the publisher has a recurring donations set (UI: #12)

![Simulator Screen Shot - iPhone X - 2019-05-15 at 16 32 01](https://user-images.githubusercontent.com/529104/57807380-0c988080-772f-11e9-878e-aaf994e91e2e.png)

PR #122, opened 2019-07-04
Added unclaimedFundsURL for when the user taps learnMore on the rewar…

Brandon-T:
Fixes: https://github.com/brave/brave-rewards-ios/issues/42

- Added unclaimedFundsURL for when the user taps learnMore on the rewards panel. 
- Added the provider to the user's name when there is one. 
- Handle "learn more" options when the user chooses it.

### DIFF
--- a/BraveRewardsUI/Wallet/WalletViewController.swift
+++ b/BraveRewardsUI/Wallet/WalletViewController.swift
@@ -158,15 +158,25 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
     publisherSummaryView.tipButton.addTarget(self, action: #selector(tappedSendTip), for: .touchUpInside)
     publisherSummaryView.monthlyTipView.addTarget(self, action: #selector(tappedMonthlyTip), for: .touchUpInside)
     
-    // TODO: Update with actual value below
+    publisherSummaryView.monthlyTipView.isHidden = true
     publisherSummaryView.monthlyTipView.batValueView.amountLabel.text = "5"
+    
     let publisherView = publisherSummaryView.publisherView
     let attentionView = publisherSummaryView.attentionView
     
     publisherView.setVerificationStatusHidden(isLocal)
+    publisherView.learnMoreTapped = { [weak self] in
+      guard let self = self, let url = URL(string: DisclaimerLinks.unclaimedFundsURL) else { return }
+      self.state.delegate?.loadNewTabWithURL(url)
+    }
+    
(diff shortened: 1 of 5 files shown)


BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -189,11 +203,19 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
         }
       }
       
+      self.state.ledger.listRecurringTips { [weak self] in
+        guard let self = self else { return }
+        guard let recurringTip = $0.filter({ $0.id == host && $0.rewardsCategory == .recurringTip }).first else { return }
```
kylehickinson: Replace `filter(_:).first` with `first(where:)`

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -158,15 +158,25 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
     publisherSummaryView.tipButton.addTarget(self, action: #selector(tappedSendTip), for: .touchUpInside)
     publisherSummaryView.monthlyTipView.addTarget(self, action: #selector(tappedMonthlyTip), for: .touchUpInside)
     
-    // TODO: Update with actual value below
+    publisherSummaryView.monthlyTipView.isHidden = true
```
kylehickinson: Is this not hidden by default already? If not carry on…

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -189,11 +203,19 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
         }
       }
       
+      self.state.ledger.listRecurringTips { [weak self] in
+        guard let self = self else { return }
+        guard let recurringTip = $0.filter({ $0.id == host && $0.rewardsCategory == .recurringTip }).first else { return }
+        
+        self.publisherSummaryView.monthlyTipView.batValueView.amountLabel.text = "\((recurringTip.contributions.first?.value ?? 0))"
```
kylehickinson: I'd say we should probably guard against showing this if we dont have an amount. So should unwrap recurringTip.contributions.first?.value before unhiding

BraveRewardsUI/Wallet/WalletViewController.swift
Brandon-T: Yeah, it wasn't hidden by default. It was always shown.

PR #123, opened 2019-07-06
Resolves #45 Added notification handler in Rewards Panel

### LINKED ISSUE #45: Show notifications on the main Rewards panel when available

danishjafri88:
Implementing #45. 
Borrowing logic from brave-core.
Ref: 
src/brave/components/brave_rewards/resources/extension/brave_rewards/components/panel.tsx
src/brave/components/brave_rewards/resources/extension/brave_rewards/_locales/en_US/messages.json
src/brave/node_modules/brave-ui/src/features/rewards/walletWrapper/index.tsx


danishjafri88:
Next steps:

- [ ] 1. Verify localisations 
- [ ] 2. Verify Icons and style
- [ ] 3. verify brave-core logic
- [ ] 4. Test with mock data.
- [ ] 5. Unit tests
- [ ] 6. Divide code into separate notification handler?

BraveRewardsUI/Wallet/Notifications/NotificationHandler.swift
```diff
@@ -0,0 +1,192 @@
...
+        addActionNotification(notification: notification)
+      default:
+        return
+      }
+    } else {
+      delegate?.hideNotifications()
+    }
+  }
+  
+  private func addActionNotification(notification: RewardsNotification) {
```
kylehickinson: `add(actionNotification:)` or `addActionNotification(_:)`

BraveRewardsUI/Wallet/Notifications/WalletActionNotificationView.swift
```diff
@@ -10,25 +10,20 @@ struct WalletActionNotification {
...
+    static let grant = Category(
       icon: UIImage(frameworkResourceNamed: "icn-grant"),
       title: Strings.NotificationTokenGrantTitle,
       action: Strings.CLAIM.uppercased()
     )
-    static let recurringTip = Category(
-      icon: UIImage(frameworkResourceNamed: "icn-contribute"),
-      title: Strings.NotificationRecurringTipTitle,
-      action: Strings.OK.uppercased()
+    static let adsLaunch = Category(
```
kylehickinson: I dont think we'll need this notification since the plan is to launch with ads

BraveRewardsUI/Wallet/Notifications/WalletNotificationView.swift
```diff
@@ -4,16 +4,63 @@
 
 import UIKit
 
+struct WalletNotification {
```
kylehickinson: If you're going to rename the other from `WalletMessageNotification` to `WalletActionNotification`, why not make this name more specific?

Also why do we now have `alerts`, `actions` and whatever this is? (Message?)

BraveRewardsUI/Wallet/Notifications/WalletNotificationView.swift
```diff
@@ -23,21 +70,72 @@ class WalletNotificationView: UIView {
...
   }
   
   @available(*, unavailable)
   required init(coder: NSCoder) {
     fatalError()
   }
+  
+  private func bodyAttributedString() -> NSAttributedString {
+    guard let notification = notification else {
+      return NSAttributedString()
```
kylehickinson: add `assertionFailure` here

BraveRewardsUI/Wallet/Notifications/NotificationHandler.swift
```diff
@@ -0,0 +1,192 @@
...
+      body = Strings.NotificationInsufficientFunds
+      category = .insufficientFunds
+    case .pendingNotEnoughFunds:
+      body = Strings.NotificationPendingNotEnoughFunds
+      category = .insufficientFunds
+    case .adsLaunch:
+      body = Strings.NotificationBraveAdsLaunchMsg
+      category = .adsLaunch
+    default:
+      loadNotification()
```
kylehickinson: Not sure what this is here for, isn't this undefined behaviour? Shouldn't we just `assertionFaliure`?

BraveRewardsUI/Wallet/Notifications/NotificationHandler.swift
```diff
@@ -0,0 +1,192 @@
...
+        category: category,
+        body: body,
+        date: date))
+    notificationView.closeButton.addTarget(self, action: #selector(tappedNotificationClose), for: .touchUpInside)
+    notificationView.actionButton.addTarget(self, action: #selector(tappedNotificationAction), for: .touchUpInside)
+    delegate?.show(notificationView: notificationView)
+  }
+  
+  @objc private func tappedNotificationClose() {
+    loadNotification()
```
kylehickinson: Honestly the fact that we're just calling `loadNotification` to clear the current one is kinda weird to me. Can we just split these out or something?

BraveRewardsUI/Wallet/Notifications/NotificationHandler.swift
```diff
@@ -0,0 +1,192 @@
...
+  }
+  
+  @objc private func tappedNotificationAction() {
+    if let notification = currrentNotification {
+      loadNotification()
```
kylehickinson: Should we be loading the next notification/clearing the current one right away?

BraveRewardsUI/Wallet/Notifications/NotificationHandler.swift
```diff
@@ -0,0 +1,192 @@
...
+      delegate?.handleAction(notification: notification)
+    }
+  }
+  
+  private func addMessageNotification(notification: RewardsNotification) {
+    var category: WalletNotification.Category = .contribute
```
kylehickinson: Why's the default value set here as contribute? If there's a reason for it, should leave documentation/comments to explain reasoning

BraveRewardsUI/Wallet/Notifications/NotificationHandler.swift
```diff
@@ -0,0 +1,192 @@
...
+      showAlertNotification(type: type, message: body)
+      return
+    }
+    
+    let date = Date(timeIntervalSince1970: notification.dateAdded)
+    let notificationView = WalletNotificationView(
+      notification: WalletNotification(
+        category: category,
+        body: body,
+        date: date))
```
kylehickinson: nit: Put trailing parenthesis on their own line

BraveRewardsUI/Wallet/Notifications/NotificationHandler.swift
```diff
@@ -0,0 +1,192 @@
...
+  @objc private func notificationAdded(_ notification: Notification) {
+    // TODO: Filter notification?
+    if currrentNotification == nil {
+      loadNotification()
+    }
+  }
+  
+  private func loadNotification() {
+    guard delegate != nil else {
+      return
```
kylehickinson: It's not on the class to worry about whether it has a delegate or not. shouldn't return just because its not set. If its vital to be set to have it work properly, it should be required in `init`

BraveRewardsUI/Wallet/Notifications/NotificationHandler.swift
```diff
@@ -0,0 +1,192 @@
...
+import Foundation
+import BraveRewards
+
+protocol NotificationHandlerDelegate: AnyObject {
+  func show(notificationView: WalletNotificationView)
+  func handleAction(notification: RewardsNotification)
+  func hideNotifications()
+}
+
+class NotificationHandler {
```
kylehickinson: Overall, I dont understand this class well…

Its acting as a receiver to turn `RewardsNotification` into one of the `Wallet*Notfication` structs (View Model), in which it then tells the delegate to show notifications (Controller) by creating a view for it (View Controller) and at the same time handle its own notification stack (Model) by clearing away things automatically.

IMO, this class shouldn't be doing all of this, and we should be picking a specific direction. Do we want it to be a "view model store" of sorts, turning `RewardsNotification`s into something that WalletViewController can read and determine what to show?

Whats stopping us of just making this class hold a list of enums that look like this:

```swift
enum WalletNotification
  case alert(AlertNotification)
  case action(ActionNotification)
  case message(MessageNotification)
}
```

On init it just looks at all ledger notifications, converts them, and adds them to its list. At the same time it subscribes to BATBraveLedgerNotificationAdded. Then you can just add simple ledger wrapper methods here that control the list (add, clear, etc.) 

And `WalletViewController` just reads that list in `viewDidLoad` and shows the first one? Then when the user clicks on something we can call `notificationStore.clear(walletNotification)` which will then call the appropriate `BATBraveLedger` method

The only thing that the notification store would still have to do is bind the `WalletViewController` to when the list updates through a delegate method.

I also don't think we should be not subscribing to changes between the point of `init` and `viewDidLoad`. That introduces a small race condition between those two points where we initially read out the notifications in the ledger and not receiving a new notification. This would be fixed by subscribing immediately on init and not forcing the `WalletViewController` to show some view when the list changes even if the view hasn't loaded yet. It'd do it itself in viewDidLoad by just reading out the notification models.

Let me know if we need further clarification around this or if I'm missing something that breaks this thought

BraveRewardsUI/Wallet/Notifications/NotificationHandler.swift
```diff
@@ -0,0 +1,192 @@
...
+  deinit {
+    stop()
+  }
+  
+  weak var currrentNotification: RewardsNotification?
+  
+  func start() {
+    // Add observer
+    stop()
+    NotificationCenter.default.addObserver(self, selector: #selector(notificationAdded(_:)), name: .init(rawValue: "BATBraveLedgerNotificationAdded"), object: nil)
```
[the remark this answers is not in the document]
danishjafri88: Did not see it :)

BraveRewardsUI/Wallet/Notifications/WalletNotificationView.swift
danishjafri88: We have three different kinds of Notifications. A normal message type, another with action and an alert type. Each has its own style and spacing.
I have made WalletNotificationView as the base view by bringing the common elements into it. Basically code inside WalletNotificationView.commonSetup() is not repeated.

Creating another subclass for Message type

BraveRewardsUI/Wallet/Notifications/WalletNotificationView.swift
danishjafri88: No more required after creating a separate class for message type

BraveRewardsUI/Wallet/Notifications/NotificationHandler.swift
danishjafri88: Clearing a notification is always followed by loading next notification. This is coupled so much that clearing a notification in itself does not have a useful case. Also the fact if clear and then set a notification, the animation is not linear but abrupt. (cases inside `setNotificationView`)

Refactoring `loadNotification` to `loadNextNotification` and creating a separate method for clearing notification and replacing it for the clear notification code in `loadNotification`

BraveRewardsUI/Wallet/Notifications/NotificationHandler.swift
danishjafri88: Created this method for contribute first and added cases later as I found them, so forgot to remove it. Updated

BraveRewardsUI/Wallet/Notifications/NotificationHandler.swift
danishjafri88: Removed, delegate must be set in `start()` for the handler to return callbacks.

BraveRewardsUI/Wallet/Notifications/NotificationHandler.swift
danishjafri88: Yeah so its basically a helper class that generates views for notifications and tells that to controller. No stack is maintained in this class, as doing so would require implementing other observers like `delete` etc from braveLedger. 

I created this class to keep WalletViewController decoupled from notification logic and also keep the number of lines in check. But I think the helper class can be further separated into more solid structures. 

I have pushed some refactors. With these changes it gets more clearer.
We have three options from it:
1. Keep as is
2. Let RewardsNotificationService forward notification and not notificationView to WalletViewController and let WalletViewController manage rest.
3. Shift all the code in RewardsNotificationService into WalletViewController.

My reason was that WalletViewController does not depend on Notifications at all. In fact the `notificationView` & `setNotificationView` in `WalletView` are also redundant(Can be simple 
 UIView). So making this class gives us the option to add/remove this feature like an addon.

I am still looking for a better way to do this, let me know what you think is better.

BraveRewardsUI/Wallet/Notifications/NotificationHandler.swift
danishjafri88: Broke it down into a callback from receiver before loading next

BraveRewardsUI/Wallet/Notifications/NotificationHandler.swift
kylehickinson: > I created this class to keep WalletViewController decoupled from notification logic

This is why I'm partially confused though because most of the logic is already decoupled because it's in `BATBraveLedger`. It has a stack, of which is managed, and you can call clear on, so all you're doing is abstracting flow logic between `RewardsNotification` and a few different view's per type of `RewardsNotification` from `WalletViewController` into a class which is only used by `WalletViewController` and is unlikely to be used anywhere else since its view specific.

And if the conversion of models to view models is what you want to pull away from WalletViewController thats fine, but the rest of the flow doesn't need to be in there (load next, close button, etc). It could just be in `WalletViewController`

If line count is a big deal you could create a new swift file that just extends `WalletViewController` and call it `WalletViewController+Notifications` or something

> My reason was that WalletViewController does not depend on Notifications at all. In fact the notificationView & setNotificationView in WalletView are also redundant(Can be simple UIView)

This doesn't have to be set in stone though, since again, these notifications are shown nowhere else.  You could just make a `UIViewController` subclass that replaces the NotificationView stuff in WalletViewController.  Then you are essentially only adding logic in `WalletViewController` on whether or not to show said controller based on non-empty notifications in the ledger and maybe getting a _did add notification_ notification which would prompt a check to notifications.

danishjafri88: Cool, I moved the code into WalletViewController.

BraveRewardsUI/Wallet/Notifications/RewardsNotificationViewBuilder.swift
```diff
@@ -0,0 +1,111 @@
...
+        date: date))
+  }
+  
+  private static func get(messageNotification: RewardsNotification) -> WalletNotificationView? {
+    var category: WalletMessageNotification.Category
+    let body: String
+    var alertType: WalletAlertNotification.Category?
+    switch messageNotification.kind {
+    case .autoContribute:
+      if let result = messageNotification.userInfo["result"] as? Int, let amount = messageNotification.userInfo["amount"] as? String {
```
kylehickinson: Use `BATResult` enum switch instead of magic numbers below by creating a `BATResult` from this line

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -14,6 +14,7 @@ protocol WalletContentView: AnyObject {
 class WalletViewController: UIViewController, RewardsSummaryProtocol {
   
   let state: RewardsState
+  weak var currentNotification: RewardsNotification?
```
kylehickinson: Why is this `weak`?

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -364,3 +343,81 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
     }
   }
 }
+
+extension WalletViewController {
+  
+  func startNotificationObserver() {
+    // Stopping as a precaution
+    stopNotificationObserver()
```
kylehickinson: Not necessary if your worry is that you'll a second observer, and since you overwrite `currentNotification` in `loadNextNotification` that's not an issue either

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -364,3 +343,81 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
...
+    // Add observer
+    NotificationCenter.default.addObserver(self, selector: #selector(notificationAdded(_:)), name: NSNotification.Name.BATBraveLedgerNotificationAdded, object: nil)
+    loadNextNotification()
+  }
+  
+  func stopNotificationObserver() {
+    NotificationCenter.default.removeObserver(self, name: NSNotification.Name.BATBraveLedgerNotificationAdded, object: nil)
```
kylehickinson: Is it possible to shorten this to `.BATBraveLedgerNotificationAdded`?

kylehickinson: Also remind me to add a `NS_SWIFT_NAME` to this in core :P

BraveRewardsUI/Wallet/Notifications/RewardsNotificationViewBuilder.swift
```diff
@@ -0,0 +1,111 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+import BraveRewards
+
+struct RewardsNotificationViewBuilder {
```
kylehickinson: This is much better! 

Would it be possible to get some unit tests for the `RewardsNotification` → `WalletActionNotification`/`WalletAlertNotification`/`WalletMessageNotification` logic?

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -97,6 +79,7 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
       navigationController?.setNavigationBarHidden(true, animated: animated)
     }
     reloadUIState()
+    startNotificationObserver()
```
kylehickinson: Is there a reason this can't be done in `init`?

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -105,6 +88,11 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
     if presentedViewController == nil {
       navigationController?.setNavigationBarHidden(false, animated: animated)
     }
+    stopNotificationObserver()
```
kylehickinson: Also is there a reason to not update the view anyways even if its not on screen? If a user pops back from settings it would just be there?

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -364,3 +343,81 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
...
+    loadNextNotification()
+  }
+  
+  @objc private func tappedNotificationAction() {
+    if let notification = currentNotification {
+      switch notification.kind {
+      case .grant, .grantAds:
+        tappedGrantsButton()
+      case .pendingNotEnoughFunds, .insufficientFunds:
+        tappedAddFunds()
```
kylehickinson: We actually _cant_ have the button here for add funds 😞

Since wallet management isn't in scope for this release despite the view already made and working, we'll have to make these message/alert notifications I guess?

BraveRewardsUI/Wallet/WalletViewController.swift
danishjafri88: Yeah, since we dont observe for deletedNotification, if any notificationView that may require an action is added and the user performs the action that deletes this notification from ledger will not hide the notificationView. The user has to manually close this notification

BraveRewardsUI/Wallet/WalletViewController.swift
danishjafri88: Since the real owner is ledger, and the ledger may remove it from its stack at any time, I found no real reason to keep it strong. If ledger deletes it then we dont have to `deleteNotification` before loading next

BraveRewardsUI/Wallet/WalletViewController.swift
danishjafri88: Added in viewdidLoad

BraveRewardsTests/RewardsNotificationViewBuilderTests.swift
```diff
@@ -0,0 +1,40 @@
...
+import BraveRewards
+import XCTest
+@testable import BraveRewardsUI
+
+class RewardsNotificationViewBuilderTests: XCTestCase {
+  
+  
+  func testBuilder() {
+    // Notifications not handled:
+    XCTAssertNil(RewardsNotificationViewBuilder.get(notification: notification(.adsLaunch)))
```
kylehickinson: Break out these tests into multiple tests (i.e. `testUnhandledNotifications`, `testActionNotifications`, `testMessageNotifications`, `testAlertNotifications`)

BraveRewardsUI/Wallet/Notifications/RewardsNotificationViewBuilder.swift
```diff
@@ -0,0 +1,115 @@
...
+      notification: WalletActionNotification(
+        category: category,
+        body: body,
+        date: date))
+  }
+  
+  private static func get(messageNotification: RewardsNotification) -> WalletNotificationView? {
+    var category: WalletMessageNotification.Category!
+    let body: String
+    var alertType: WalletAlertNotification.Category?
```
kylehickinson: You've split these out into multiple types and yet in order to get an alert notification you have to call `get(messageNotification: RewardsNotification)`? Isn't there a better way to set this up so there's only 1 type. Is the message + alert notification the same other than the image?

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -427,3 +405,71 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
...
+    if let currentNotification = currentNotification {
+      state.ledger.clearNotification(currentNotification)
+      self.currentNotification = nil
+    }
+  }
+  
+  private func loadNextNotification() {
+    //.failedContribution, .adsLaunch, .invalid & .backupWallet are not supported right now
+    // TODO: Verify that notifications are ordered & they are displayed in FIFO.
+    let ignoredRewardNotificationTypes: [RewardsNotification.Kind] = [.failedContribution,
```
kylehickinson: We have this list of ignored notifications in 2 places now, is it possible to consolidate it somehow?

danishjafri88: I can only see this. Can you mark the other place?

kylehickinson: https://github.com/brave/brave-rewards-ios/pull/123/files#diff-1ed66972ddc2dd1522f79ac2731e9a61R17

Implicitly here

BraveRewardsUI/Wallet/Notifications/RewardsNotificationViewBuilder.swift
danishjafri88: Auto-Contribute notifications can either have a message notification or an alert notification, depending on `BATResult`. Message notification and alert notification have different UI.

BraveRewardsExample/ViewController.swift
```diff
@@ -106,6 +130,12 @@ class ViewController: UIViewController {
     
     setupRewards()
     braveRewardsPanelButton.setImage(RewardsPanelController.batLogoImage, for: .normal)
+//    DispatchQueue.main.asyncAfter(deadline: .now() + 15) {
```
kylehickinson: Remove?

BraveRewardsTests/RewardsNotificationViewBuilderTests.swift
```diff
@@ -0,0 +1,56 @@
...
+
+import Foundation
+import BraveRewards
+import XCTest
+@testable import BraveRewardsUI
+
+class RewardsNotificationViewBuilderTests: XCTestCase {
+  
+  
+  func testBuilder() {
```
kylehickinson: I'm not sure why you have this?

BraveRewardsTests/RewardsNotificationViewBuilderTests.swift
```diff
@@ -0,0 +1,56 @@
...
+   
+    testUnHandledNotifications()
+    // Wanted cases
+    testAlertNotification()
+    testMessageNotification()
+    testActionNotifiication()
+  }
+  
+  func testUnHandledNotifications() {
```
kylehickinson: `UnHandled` → `Unhandled`

BraveRewardsTests/RewardsNotificationViewBuilderTests.swift
```diff
@@ -0,0 +1,56 @@
...
+    // Notifications not handled:
+    XCTAssertNil(RewardsNotificationViewBuilder.get(notification: notification(.adsLaunch)))
+    
+    XCTAssertNil(RewardsNotificationViewBuilder.get(notification: notification(.backupWallet)))
+
```
kylehickinson: Remove whitespace

BraveRewardsExample/ViewController.swift
danishjafri88: Ah.. was testing postnotification forgot to remove

BraveRewardsUI/Wallet/WalletViewController.swift
danishjafri88: well that list is used to filter unwanted builder calls, and first check to stop handling certain notifications. Builder can still have the code to create view for unwanted notifications. Removed filter for now

danishjafri88:
> Incorrect UI design implementation:
> 
> Design	Implementation
> image	image
> image	image
> Not sure if you talked to design about these changes?

Nope I actually took que from desktop. They dont have a button for these types. Rather they had the UI as right now. Will update

kylehickinson:
Would rather just have consistency across platforms so I'm fine with them not matching design if design's fine with it.

danishjafri88:
> Incorrect UI design implementation:
> 
> Design	Implementation
> image	image
> image	image
> Not sure if you talked to design about these changes?

Updated designs after talking with @jamesmudgett, waiting for a final go ahead. 
For reference:

Simulator Screen Shot - iPhone X - 2019-07-17 at 01 22 41
Simulator Screen Shot - iPhone X - 2019-07-17 at 01 22 43
Simulator Screen Shot - iPhone X - 2019-07-17 at 01 22 44
Simulator Screen Shot - iPhone X - 2019-07-17 at 01 22 45
Simulator Screen Shot - iPhone X - 2019-07-17 at 01 22 46
Simulator Screen Shot - iPhone X - 2019-07-17 at 01 22 47
Simulator Screen Shot - iPhone X - 2019-07-17 at 01 25 06

BraveRewardsTests/RewardsNotificationViewBuilderTests.swift
```diff
@@ -0,0 +1,42 @@
...
+    XCTAssertNil(RewardsNotificationViewBuilder.get(notification: notification(.failedContribution)))
+    XCTAssertNil(RewardsNotificationViewBuilder.get(notification: notification(.invalid)))
+  }
+  
+  func testActionNotifiication() {
+    XCTAssert(RewardsNotificationViewBuilder.get(notification: notification(.grant)) is WalletActionNotificationView)
+  }
+  
+  func testMessageNotification() {
+    XCTAssert(RewardsNotificationViewBuilder.get(notification: notification(.tipsProcessed)) is WalletActionNotificationView)
```
kylehickinson: Probably should just move the assertions inside this test into `testActionNotifiication` since there's no "message notification" anymore

danishjafri88: Oops, thanks for noticing. Updated now

PR #130, opened 2019-07-08
Update library with new favicon fetching API

kylehickinson:
Adds support for new favicon fetching API changes

[Library changset](https://github.com/brave/brave-core/commits/master/vendor/brave-ios):
- Generate default value assignments for Obj-C mojo gen https://github.com/brave/brave-core/pull/2879
  - Also forces default values onto CoreData models instead of being nullable
  - Fixes tests that created models without a publisher dependency (now that reverse relationships are nonnull)
- Add favicon fetching handler for brave-ios to implement https://github.com/brave/brave-core/pull/2866
- Removes non public functions https://github.com/brave/brave-core/pull/2863
- Removes OnMediaStart and OnMediaStop as it's not used https://github.com/brave/brave-core/pull/2862
- Removes AddRecurringPayment as it's not used https://github.com/brave/brave-core/pull/2864


BraveRewardsExample/ViewController.swift
```diff
@@ -163,9 +173,9 @@ extension ViewController: RewardsDataSource {
     return url.host
   }
   
-  func retrieveFavicon(with url: URL, completion: @escaping (FaviconData?) -> Void) {
+  func retrieveFavicon(for pageURL: URL, faviconURL: URL?, completion: @escaping (FaviconData?) -> Void) {
```
iccub: What I'm supposed to do here in `brave-ios`?
Get favicon by faviconURL, or by pageURL if faviconURL is nil?

kylehickinson: We can do either, I really only updated it to include both because it was confusing before which URL was sent and brave-ios could technically do both :) Now we have a choice

BraveRewardsExample/ViewController.swift
```diff
@@ -147,6 +147,16 @@ class ViewController: UIViewController {
   }
 }
 
+extension ViewController: BraveRewardsDelegate {
+  func faviconURL(fromPageURL pageURL: URL, completion: @escaping (URL?) -> Void) {
```
iccub: This is never called

kylehickinson: I think this is because we never use `GetActivityInfoFromUrl` or whatever that method is called. Therefore no saveVisitInternal happens with a faviconUrl passed in

BraveRewardsUI/Settings/Auto-Contribute/Details/AutoContributeDetailsViewController.swift
```diff
@@ -319,7 +319,7 @@ extension AutoContributeDetailViewController: UITableViewDataSource, UITableView
       cell.selectionStyle = .none
       
       if let url = URL(string: publisher.url) {
-        state.dataSource?.retrieveFavicon(with: url) { data in
+        state.dataSource?.retrieveFavicon(for: url, faviconURL: URL(string: publisher.faviconUrl)) { data in
```
iccub: can `faviconUrl` be set to `clear` string here? or will it be nil when `clear` param is passed?
I remember desktop doing having that `clear` property

kylehickinson: `clear` is no longer an option here, it'll just be nil. `clear` is taken care of in the `-[BATLedgerDatabase insertOrUpdatePublisherInfo:completion]` call and replaced with an empty string (which will then become nil here)

kylehickinson: As seen here: https://github.com/brave/brave-core/pull/2866/files#diff-22cd939021706f39a7c4ec71cb224917R154

BraveRewardsUI/Tipping/TippingViewController.swift
```diff
@@ -99,8 +99,10 @@ class TippingViewController: UIViewController, UIViewControllerTransitioningDele
         self.tippingView.overviewView.headerView.image = image
       })
       
-      if let dataSource = self.state.dataSource, let favIconURL = self.state.faviconURL {
-        dataSource.retrieveFavicon(with: favIconURL, completion: { favIconData in
+      if let dataSource = self.state.dataSource,
+        let pageURL = URL(string: self.publisherInfo.url),
+        let favIconURL = self.state.faviconURL {
+        dataSource.retrieveFavicon(for: pageURL, faviconURL: favIconURL, completion: { favIconData in
```
iccub: replace `favIcon` with `favicon` for naming consistency

kylehickinson:
Will merge post-demo

Issue #129, opened 2019-07-08
Verified publisher status is missing


kylehickinson:
Should say verified or unverified publisher under the name. Seems to be missing?

![Simulator Screen Shot - iPhone 8 - 2019-07-08 at 17 32 15](https://user-images.githubusercontent.com/529104/60844142-715bde80-a1a6-11e9-8161-fea7393180de.png)

Issue #127, opened 2019-07-08
Show BAT rates on recurring tip amount selector screen


kylehickinson:
This should read "1.0 BAT (X USD)" instead of just 1.0/5.0/10.0 similar to other BAT selection screens shown in settings

![Simulator Screen Shot - iPhone 8 - 2019-07-08 at 17 16 48](https://user-images.githubusercontent.com/529104/60843288-30fb6100-a1a4-11e9-9f57-9e5cf9fbc6f0.png)

Issue #126, opened 2019-07-08
Add ability to remove recurring tip


kylehickinson:
When the user is on this screen after tapping on the publishers recurring tip, they should be able to choose 0 BAT to remove it

Also we should give a title to this list

![Simulator Screen Shot - iPhone 8 - 2019-07-08 at 17 16 48](https://user-images.githubusercontent.com/529104/60843288-30fb6100-a1a4-11e9-9f57-9e5cf9fbc6f0.png)

PR #131, opened 2019-07-10
Tipping selection, remove recurring tips

Brandon-T:
Updated tipping selection to maintain state between screens when a tip is selected.
Updated tipping selection to allow the user to remove a tip by selecting "0 BAT".
Updated tipping selection screen to show the currency and have a title.

Fixes https://github.com/brave/brave-rewards-ios/issues/126
Fixes https://github.com/brave/brave-rewards-ios/issues/127
Fixes https://github.com/brave/brave-rewards-ios/issues/129
Fixes https://github.com/brave/brave-rewards-ios/issues/134

### DIFF
--- a/BraveRewardsExample/ViewController.swift
+++ b/BraveRewardsExample/ViewController.swift
@@ -40,6 +40,13 @@ class UIMockLedger: BraveLedger {
     }
   }
   
+  override var balance: Balance? {
+    return Balance().then {
+      $0.total = 30.0
+      $0.rates = ["USD": 0.3]
+    }
+  }
+  
   override func balanceReport(for month: ActivityMonth, year: Int32) -> BalanceReportInfo {
     // Returns the same values for all months at the moment.
     // Could expand this to play with different values based on month or year.
--- a/BraveRewardsUI/Common/Button.swift
+++ b/BraveRewardsUI/Common/Button.swift
@@ -80,7 +80,6 @@ class Button: UIButton {
             loaderView.removeFromSuperview()
           }
         }
-        break
       }
     }
   }
--- a/BraveRewardsUI/Tipping/TippingSelectionViewController.swift
+++ b/BraveRewardsUI/Tipping/TippingSelectionViewController.swift
@@ -1,0 +1,65 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+import UIKit
+import BraveRewards
+
+class TippingSelectionViewController: OptionsSelectionViewController {
+  
+  private let state: RewardsState?
+  
+  // Hide the super constructor with `DisplayableOption`
+  // in favour of our constructor with `BatValue`
+  private override init(options: [DisplayableOption],
+                        selectedOptionIndex: Int = 0,
+                        optionSelected: @escaping (_ selectedIndex: Int) -> Void) {
+    self.state = nil
+    super.init(options: options, selectedOptionIndex: selectedOptionIndex, optionSelected: optionSelected)
+  }
+  
+  init(state: RewardsState, options: [BATValue],
+       selectedOptionIndex: Int = 0,
+       optionSelected: @escaping (_ selectedIndex: Int) -> Void) {
+    self.state = state
+    super.init(options: options, selectedOptionIndex: selectedOptionIndex, optionSelected: optionSelected)
+  }
+  
+  override func viewDidLoad() {
+    super.viewDidLoad()
+    
+    self.title = "Recurring Tips"
+  }
+  
+  override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+    let cell = tableView.dequeueReusableCell(withIdentifier: "OptionCell", for: indexPath)
+    cell.textLabel?.text = options[indexPath.row].displayString
+    cell.textLabel?.font = .systemFont(ofSize: 14.0)
+    cell.textLabel?.textColor = Colors.grey100
+    cell.textLabel?.numberOfLines = 0
+    cell.accessoryType = selectedOptionIndex == indexPath.row ? .checkmark : .none
+    
+    guard let options = options as? [BATValue] else { return cell }
+    let displayString = options[indexPath.row].displayString
+    let dollarAmount = state?.ledger.dollarStringForBATAmount(options[indexPath.row].doubleValue) ?? ""
+    
+    let attributedText = NSMutableAttributedString(string: displayString, attributes: [
+      .foregroundColor: Colors.grey100,
+      .font: UIFont.systemFont(ofSize: 14.0, weight: .medium)
+    ])
+    
+    attributedText.append(NSAttributedString(string: " BAT", attributes: [
+      .foregroundColor: Colors.grey200,
+      .font: UIFont.systemFont(ofSize: 12.0)
+    ]))
+    
+    attributedText.append(NSAttributedString(string: " (\(dollarAmount))", attributes: [
+      .foregroundColor: Colors.grey200,
+      .font: UIFont.systemFont(ofSize: 10.0)
+    ]))
+    
+    cell.textLabel?.attributedText = attributedText
+    return cell
+  }
+}
--- a/BraveRewardsUI/Tipping/TippingSelectionViewController.swift
+++ b/BraveRewardsUI/Tipping/TippingSelectionViewController.swift
@@ -6,43 +6,28 @@ import Foundation
 import UIKit
 import BraveRewards
 
-class TippingSelectionViewController: OptionsSelectionViewController {
+class BATValueOptionsSelectionViewController: OptionsSelectionViewController<BATValue> {
   
-  private let state: RewardsState?
+  private let ledger: BraveLedger?
   
-  // Hide the super constructor with `DisplayableOption`
-  // in favour of our constructor with `BatValue`
-  private override init(options: [DisplayableOption],
-                        selectedOptionIndex: Int = 0,
-                        optionSelected: @escaping (_ selectedIndex: Int) -> Void) {
-    self.state = nil
-    super.init(options: options, selectedOptionIndex: selectedOptionIndex, optionSelected: optionSelected)
-  }
-  
-  init(state: RewardsState, options: [BATValue],
+  init(ledger: BraveLedger, options: [BATValue],
        selectedOptionIndex: Int = 0,
        optionSelected: @escaping (_ selectedIndex: Int) -> Void) {
-    self.state = state
+    self.ledger = ledger
     super.init(options: options, selectedOptionIndex: selectedOptionIndex, optionSelected: optionSelected)
   }
   
   override func viewDidLoad() {
     super.viewDidLoad()
     
-    self.title = "Recurring Tips"
+    self.title = Strings.RecurringTipTitle
   }
   
   override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-    let cell = tableView.dequeueReusableCell(withIdentifier: "OptionCell", for: indexPath)
-    cell.textLabel?.text = options[indexPath.row].displayString
-    cell.textLabel?.font = .systemFont(ofSize: 14.0)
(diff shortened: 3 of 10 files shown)


BraveRewardsUI/Tipping/TippingSelectionViewController.swift
```diff
@@ -0,0 +1,65 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+import UIKit
+import BraveRewards
+
+class TippingSelectionViewController: OptionsSelectionViewController {
```
kylehickinson: A subclass for this one use case is probably over the top, and could have just copied this:

https://github.com/brave/brave-rewards-ios/blob/4fcd743924e9573547aa2bfb3d44cecf5f766d3d/BraveRewardsUI/Settings/Auto-Contribute/Settings/AutoContributeSettingsViewController.swift#L97-L118

If we're going to do this just to get custom appearance for the text, we should just make it generic like `BATSelectionViewController` and use it in auto-contribute settings as well

We'll have to make `OptionsSelectionViewController` deceleration use generics in this case like 

```swift
class OptionsSelectionViewController<OptionType: DisplayableOption>: UIViewController` {
```

Then here you can do:

```swift
class BATValueOptionsSelectionViewController: OptionsSelectionViewController<BATValue> {
```

Then when you make a new designated initializer it won't complain that the types are wrong

If we can't get that part working properly for now or has more issues, lets just scrap it and do the non-styled version from Auto-Contribute Settings for now

BraveRewardsUI/Tipping/TippingSelectionViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+  
+  private let state: RewardsState?
```
kylehickinson: Once changes are made to `OptionsSelectionViewController` as explained above, you won't need a private init and can make this non-optional, since its assigned in init

BraveRewardsUI/Tipping/TippingSelectionViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+  }
+  
+  override func viewDidLoad() {
+    super.viewDidLoad()
+    
+    self.title = "Recurring Tips"
+  }
+  
+  override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+    let cell = tableView.dequeueReusableCell(withIdentifier: "OptionCell", for: indexPath)
```
kylehickinson: Just call super cell for path? Then you already get the dequeued cell without using assumed cell identifier

BraveRewardsUI/Tipping/TippingSelectionViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+       selectedOptionIndex: Int = 0,
+       optionSelected: @escaping (_ selectedIndex: Int) -> Void) {
+    self.state = state
+    super.init(options: options, selectedOptionIndex: selectedOptionIndex, optionSelected: optionSelected)
+  }
+  
+  override func viewDidLoad() {
+    super.viewDidLoad()
+    
+    self.title = "Recurring Tips"
```
kylehickinson: 1. Should be localized/placed in `Strings.swift`
2. Lets go with singular `Recurring Tip` instead

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -159,7 +160,7 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
     publisherSummaryView.monthlyTipView.addTarget(self, action: #selector(tappedMonthlyTip), for: .touchUpInside)
     
     publisherSummaryView.monthlyTipView.isHidden = true
-    publisherSummaryView.monthlyTipView.batValueView.amountLabel.text = "5"
+    publisherSummaryView.monthlyTipView.batValueView.amountLabel.text = "\(Int(self.recurringTipAmount))"
```
kylehickinson: This is set way later down the line already, is this needed? Wont it always be 0

BraveRewardsUI/Tipping/TippingSelectionViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+  // Hide the super constructor with `DisplayableOption`
+  // in favour of our constructor with `BatValue`
+  private override init(options: [DisplayableOption],
+                        selectedOptionIndex: Int = 0,
+                        optionSelected: @escaping (_ selectedIndex: Int) -> Void) {
+    self.state = nil
+    super.init(options: options, selectedOptionIndex: selectedOptionIndex, optionSelected: optionSelected)
+  }
+  
+  init(state: RewardsState, options: [BATValue],
```
kylehickinson: I don't _love_ sending in `RewardsState` in here, and I know its needed for rates, but its fine for now. Though, do we need it to be a `RewardsState` or can we just send in a `BraveLedger` instead

BraveRewardsUI/Tipping/TippingSelectionViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+  
+  override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+    let cell = tableView.dequeueReusableCell(withIdentifier: "OptionCell", for: indexPath)
+    cell.textLabel?.text = options[indexPath.row].displayString
+    cell.textLabel?.font = .systemFont(ofSize: 14.0)
+    cell.textLabel?.textColor = Colors.grey100
+    cell.textLabel?.numberOfLines = 0
+    cell.accessoryType = selectedOptionIndex == indexPath.row ? .checkmark : .none
+    
+    guard let options = options as? [BATValue] else { return cell }
```
kylehickinson: With above changes we won't need this anymore

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -317,23 +319,37 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
...
+        guard let self = self else { return }
+        
+        self.recurringTipAmount = options[optionIndex].doubleValue
+        
+        self.navigationController?.popToViewController(self, animated: true)
+        // swiftlint:ignore:next
+        self.publisherSummaryView.monthlyTipView.batValueView.amountLabel.text = options[safe: optionIndex]?.displayString ?? options[0].displayString
+        
+        // The user decided to remove an existing tip.
+        if optionIndex == 0 && selectedIndex > 0 {
```
kylehickinson: Probably safer to check if the option (`BATValue`)'s value is 0 instead of just checking the option index. 

I don't think the second check to `selectedIndex` is needed either since choosing zero regardless of what you had previously selected should remove the tip, considering you can't get to this screen without having a non-zero tip

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -218,9 +198,16 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
...
         publisherView.checkAgainButton.isLoading = true
         
-        self.state.ledger.refreshPublisher(withId: host, completion: { _ in
+        self.state.ledger.refreshPublisher(withId: host, completion: { [weak self] didRefresh in
+          guard let self = self else { return }
+          
           publisherView.checkAgainButton.isLoading = false
-          publisherView.checkAgainButton.isHidden = true
+          
+          if didRefresh {
```
kylehickinson: The bool that comes back from this is not `didRefresh` its `isVerified`, so not sure if this logic is correct?

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -247,7 +234,40 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
         attentionView.valueLabel.text = "\(info.percent)%"
       }
     }
-
+  }
+  
+  private func updatePublisherView(completion: (() -> Void)?) {
+    guard let host = state.url.host else { return }
+    self.state.ledger.publisherInfo(forId: host) { [weak self] info in
```
kylehickinson: `publisherInfo(forId:)` is not actually asynchronous, and the block is non-escaping on the Obj-C++ side, so you can technically remove the `completion` block

I'm going to add `NS_NOESCAPE` to blocks to the Obj-C++ side that aren't actually escaping or even update said APIs to not use completion blocks if I can so its more obvious that we can use the results of the block synchronously

BraveRewardsUI/Tipping/TippingSelectionViewController.swift
Brandon-T: Fixed.

BraveRewardsUI/Tipping/TippingSelectionViewController.swift
Brandon-T: Fixed.

BraveRewardsUI/Tipping/TippingSelectionViewController.swift
Brandon-T: Fixed.

BraveRewardsUI/Tipping/TippingSelectionViewController.swift
Brandon-T: Fixed.

BraveRewardsUI/Wallet/WalletViewController.swift

BraveRewardsUI/Wallet/WalletViewController.swift

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -208,17 +213,27 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
...
-        publisherView.checkAgainButton.isLoading = true
-        
-        self.state.ledger.refreshPublisher(withId: host, completion: { _ in
-          publisherView.checkAgainButton.isLoading = false
-          publisherView.checkAgainButton.isHidden = true
+        self?.state.ledger.refreshPublisher(withId: host, completion: { isVerified in
+          
+          if isVerified {
+            publisherView.checkAgainButton.isLoading = false
+            publisherView.checkAgainButton.isHidden = true
```
kylehickinson: dont you need to call `publisherView.setVerified(isVerified)` here?

BraveRewardsUI/Wallet/WalletViewController.swift

Brandon-T:
> Not seeing any click-state for the "Check Again…" button and no loader when clicked:

I added:

```swift
Button(type: .system).then {
    $0.loaderPlacement = .system    //Uses a transition animation to hide labels and images and place the loader visible.
}
```

BraveRewardsUI/Common/Button.swift
```diff
@@ -80,7 +86,33 @@ class Button: UIButton {
             loaderView.removeFromSuperview()
           }
         }
-        break
+      case .system:
+        let isLoading = self.isLoading
+        let colour = self.titleColor(for: .normal)
```
kylehickinson: Won't this be clear and the image below be nil upon any subsequent loading toggles?

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -208,17 +213,31 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
...
-          publisherView.checkAgainButton.isHidden = true
+        self?.state.ledger.refreshPublisher(withId: host, completion: { isVerified in
+          
+          let updateStates = {
+            publisherView.setCheckAgainIsLoading(false)
+            publisherView.checkAgainButton.isHidden = true
+            publisherView.setVerified(isVerified)
+          }
+          
+          //Fixes brave-rewards-ios/issues/134
```
kylehickinson: Instead of reffing issue just mention copying desktop

BraveRewardsExample/ViewController.swift
```diff
@@ -88,7 +95,7 @@ class ViewController: UIViewController {
   
   var rewards: BraveRewards!
   
-  private static let testPublisherURL = "https://3zsistemi.si" //"https://bumpsmack.com"
+  private static let testPublisherURL = "https://google.ca" //"https://bumpsmack.com"
```
kylehickinson: Sorry can we revert this one

Issue #137, opened 2019-07-16
Use new ledger observers within the brave panel UI


kylehickinson:
Ref #90 

Observers added here: https://github.com/brave/brave-core/pull/2922 (will require lib update on `brave-rewards-ios` side)

Should be used appropriately throughout the UI

PR #144, opened 2019-07-30
Added observers for ledger events

danishjafri88:
#137  Added Observers

### DIFF
--- a/BraveRewardsUI/Settings/Auto-Contribute/Details/AutoContributeDetailsViewController.swift
+++ b/BraveRewardsUI/Settings/Auto-Contribute/Details/AutoContributeDetailsViewController.swift
@@ -6,7 +6,7 @@ import UIKit
 import BraveRewards
 
 class AutoContributeDetailViewController: UIViewController {
-  
+  private var ledgerObserver: LedgerObserver
   private var contentView: View {
     return view as! View // swiftlint:disable:this force_cast
   }
@@ -19,7 +19,10 @@ class AutoContributeDetailViewController: UIViewController {
   
   init(state: RewardsState) {
     self.state = state
+    ledgerObserver = LedgerObserver(ledger: state.ledger)
+    state.ledger.add(ledgerObserver)
     super.init(nibName: nil, bundle: nil)
+    setupLedgerObservers()
   }
   
   @available(*, unavailable)
@@ -33,7 +36,6 @@ class AutoContributeDetailViewController: UIViewController {
   
   override func viewDidLoad() {
     super.viewDidLoad()
-    
     contentView.tableView.delegate = self
     contentView.tableView.dataSource = self
     
@@ -193,8 +195,6 @@ extension AutoContributeDetailViewController: UITableViewDataSource, UITableView
         let alert = UIAlertController(title: nil, message: nil, preferredStyle: .actionSheet)
         alert.addAction(UIAlertAction(title: String(format: Strings.AutoContributeRestoreExcludedSites, numberOfExcludedSites), style: .default, handler: { _ in
           self.state.ledger.restoreAllExcludedPublishers()
-          // TODO: Animate this update
-          self.contentView.tableView.reloadData()
         }))
         alert.addAction(UIAlertAction(title: Strings.Cancel, style: .cancel, handler: nil))
         present(alert, animated: true)
@@ -258,9 +258,6 @@ extension AutoContributeDetailViewController: UITableViewDataSource, UITableView
     guard Section(rawValue: indexPath.section) == .contributions else { return }
     if let publisher = publishers[safe: indexPath.row] {
       state.ledger.updatePublisherExclusionState(withId: publisher.id, state: .excluded)
-      publishers.remove(at: indexPath.row)
-      // TODO: Animate this update
-      tableView.reloadData()
     }
   }
   
@@ -279,30 +276,33 @@ extension AutoContributeDetailViewController: UITableViewDataSource, UITableView
       cell.label.numberOfLines = 0
       cell.accessoryLabel?.textColor = Colors.grey100
       cell.accessoryLabel?.font = SettingsUX.bodyFont
+      cell.accessoryType = .none
+      cell.selectionStyle = .none
       switch row {
       case .settings:
         cell.label.text = Strings.Settings
         cell.imageView?.image = UIImage(frameworkResourceNamed: "settings").alwaysTemplate
         cell.imageView?.tintColor = BraveUX.autoContributeTintColor
         cell.accessoryType = .disclosureIndicator
+        cell.selectionStyle = .default
       case .monthlyPayment:
         cell.label.text = Strings.AutoContributeMonthlyPayment
         cell.accessoryType = .disclosureIndicator
         if let dollarAmount = state.ledger.dollarStringForBATAmount(state.ledger.contributionAmount) {
           cell.accessoryLabel?.text = "\(state.ledger.contributionAmount) BAT (\(dollarAmount))"
         }
+        cell.selectionStyle = .default
       case .nextContribution:
         cell.label.text = Strings.AutoContributeNextDate
         cell.accessoryView = nextContributionDateView
-        cell.selectionStyle = .none
       case .supportedSites:
         cell.label.text = Strings.AutoContributeSupportedSites
         cell.accessoryLabel?.attributedText = totalSitesAttributedString(from: publishers.count)
-        cell.selectionStyle = .none
       case .excludedSites:
         let numberOfExcludedSites = state.ledger.numberOfExcludedPublishers
         cell.label.text = String(format: Strings.AutoContributeRestoreExcludedSites, numberOfExcludedSites)
         cell.label.textColor = Colors.blurple400
+        cell.selectionStyle = .default
       }
       return cell
     case .contributions:
@@ -320,7 +320,7 @@ extension AutoContributeDetailViewController: UITableViewDataSource, UITableView
       
       if let url = URL(string: publisher.url) {
         state.dataSource?.retrieveFavicon(for: url, faviconURL: URL(string: publisher.faviconUrl)) { data in
-          cell.siteImageView.image = data?.image
+          cell.siteImageView.image = data?.image ?? UIImage(frameworkResourceNamed: "defaultFavicon")
           cell.siteImageView.backgroundColor = data?.backgroundColor
         }
       }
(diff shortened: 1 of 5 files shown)


BraveRewardsUI/Settings/Auto-Contribute/Details/AutoContributeDetailsViewController.swift
```diff
@@ -6,7 +6,7 @@ import UIKit
 import BraveRewards
 
 class AutoContributeDetailViewController: UIViewController {
-  
+  private var ledgerObs: LedgerObserver
```
kylehickinson: Rather not short-form it :)

BraveRewardsUI/Settings/Auto-Contribute/Details/AutoContributeDetailsViewController.swift
```diff
@@ -31,9 +33,13 @@ class AutoContributeDetailViewController: UIViewController {
     self.view = View()
   }
   
+  deinit {
+    state.ledger.remove(ledgerObs)
```
kylehickinson: Not necessary

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
```diff
@@ -6,7 +6,7 @@ import UIKit
 import BraveRewards
 
 class TipsDetailViewController: UIViewController {
-  
+  private var ledgerObs: LedgerObserver
```
kylehickinson: Feedback same as auto-contribute file

BraveRewardsUI/Settings/Auto-Contribute/Details/AutoContributeDetailsViewController.swift
```diff
@@ -382,3 +386,47 @@ extension AutoContributeDetailViewController {
     }
   }
 }
+
+/// Ledger Observers
+extension AutoContributeDetailViewController {
+  func setObservers() {
```
kylehickinson: Assuming this is only ever done once, lets call it `setupLedgerObservers`

Also lets have this setup in `init`, and add `guard viewIsLoaded else { return }` before any part of the methods require the view constructed.  In this case below its going to be at the top

BraveRewardsUI/Settings/Auto-Contribute/Details/AutoContributeCell.swift
```diff
@@ -48,6 +48,8 @@ class AutoContributeCell: UITableViewCell, TableViewReusable {
     siteStackView.addArrangedSubview(siteImageView)
     siteStackView.addArrangedSubview(siteNameLabel)
     
+    siteImageView.snp.removeConstraints()
```
kylehickinson: What's this for?

BraveRewardsUI/Settings/Auto-Contribute/Details/AutoContributeDetailsViewController.swift
```diff
@@ -382,3 +386,47 @@ extension AutoContributeDetailViewController {
...
+    ledgerObs.excludedSitesChanged = { key, exclude -> Void in
```
kylehickinson: `[weak self]` since we are holding a ref to `ledgerObs` in self

BraveRewardsUI/Settings/Auto-Contribute/Details/AutoContributeCell.swift
danishjafri88: siteImageView's constructor takes a height which it uses to set a height constraint. But the imageview is placed inside a stackview and the stackview is inside a self sizing UITableviewcell. When it's drawn because of conflicting constraints, causes jerks. Removing this makes imageview draw relative to the stackview.

kylehickinson: Ahhh okay cool didn't notice it was a custom image view

BraveRewardsUI/Tipping/TippingViewController.swift
```diff
@@ -25,17 +25,18 @@ enum PublisherMediaType: String {
...
   
   init(state: RewardsState, publisherInfo: PublisherInfo) {
     self.state = state
     self.publisherInfo = publisherInfo
-    
+    ledgerObserver = LedgerObserver(ledger: state.ledger)
+    state.ledger.add(ledgerObserver)
     super.init(nibName: nil, bundle: nil)
-    
+    setupLedgerObservers()
```
kylehickinson: We're going to handle showing confirmation based on callbacks added in #143. Unfortunately we won't want to wait for reconcile or observe recurring tips being added since we won't know when a reconcile will complete (could never happen) and 2. technically but likley never going to happen: other recurring tips could be added that are unrelated

Whether or not the tip goes out instantly, or its just added to pending contributions table, is dependent on a bunch of factors

So likely just going to take out all observer stuff in `TippingViewController`

danishjafri88:
> Still crashing when removing multiple auto-contribute entries:
> 
> ```
> Invalid update: invalid number of rows in section 0.  The number of rows contained in an existing section after the update (5) must be equal to the number of rows contained in that section before the update (5), plus or minus the number of rows inserted or deleted from that section (1 inserted, 0 deleted) and plus or minus the number of rows moved into or out of that section (0 moved in, 0 moved out).
> ```
> 
> My steps that caused the crash:
> 
> 1. Restore All publishers (2), causing the restore all publishers row to go away
> 2. Scroll down (top section is out of view), exclude 1 publisher
> 3. Exclude a second publisher
> 
> Crash line: https://github.com/brave/brave-rewards-ios/pull/144/files#diff-33d19beafb04380e3ae8944a9e34df8eR420
> 
> **Note**: I rebased this branch so make sure you delete your local branch before pulling

Fixed and pushed.

danishjafri88:
> Sorry its still weirdly glitchy when removing, lets just dump the animations and just `reloadData` for MVP. We can revisit later:
> 
> [rewards-ac-animation.mov.zip](https://github.com/brave/brave-rewards-ios/files/3576463/rewards-ac-animation.mov.zip)

Yeah, its because we are updating multiple things: 1 Delete, 1 insert (Different sections) and reload visible cells in section 1. Moving reloadData for now.

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
```diff
@@ -279,4 +283,26 @@ extension TipsDetailViewController {
...
+        $0.category == RewardsCategory.recurringTip.rawValue &&
+        $0.id == key
+      })
+      if oldCount > self.tipsList.count {
+        var sections: IndexSet = [Section.tips.rawValue]
+        if tableView.numberOfRows(inSection: Section.summary.rawValue) == 2 &&
+          !self.tipsList.contains { $0.rewardsCategory == .recurringTip } {
+          sections.insert(Section.summary.rawValue)
+        }
+        self.tipsView.tableView.reloadSections(sections, with: .automatic)
```
kylehickinson: still animating this, I think the whole section here can be replaced right?

BraveRewardsUI/Settings/Auto-Contribute/Details/AutoContributeDetailsViewController.swift
```diff
@@ -382,3 +382,28 @@ extension AutoContributeDetailViewController {
...
+    ledgerObserver.excludedSitesChanged = { [weak self] key, exclude -> Void in
+      guard let self = self, self.isViewLoaded else {
+        return
+      }
+      let tableView = self.contentView.tableView
+      switch exclude {
+      case .all:
+        tableView.reloadData()
+      case .excluded:
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.3, execute: {
```
kylehickinson: Is this delay still needed without animations?

BraveRewardsUI/Settings/Wallet/WalletDetailsViewController.swift
```diff
@@ -52,4 +54,30 @@ class WalletDetailsViewController: UIViewController, RewardsSummaryProtocol {
...
     present(container, animated: true)
   }
+  
+  func setupLedgerObservers() {
+    ledgerObserver.balanceReportUpdated = { [weak self] in
+      guard let self = self, self.isViewLoaded else {
+        return
+      }
+      let rows = self.summaryRows.map({ row -> RowView in
+        row.isHidden = true
```
kylehickinson: Why are you setting them all to hidden here?

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -523,4 +512,42 @@ extension WalletViewController {
...
+    ledgerObserver.activityRemoved = { [weak self] publisherKey in
+      guard let self = self, publisherKey == self.publisher?.id else { return }
+      self.fetchPublisherActivity()
+    }
+    ledgerObserver.balanceReportUpdated = { [weak self] in
+      guard let self = self, self.isViewLoaded else {
+        return
+      }
+      let rows = self.summaryRows.map({ row -> RowView in
+        row.isHidden = true
```
kylehickinson: Same question as above

BraveRewardsUI/Settings/Wallet/WalletDetailsViewController.swift
danishjafri88: To animate an item in stack view we have to toggle isHidden

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
danishjafri88: This is a different controller, should I remove animation here too?

kylehickinson: Yeah lets just do all of it

BraveRewardsUI/Settings/Wallet/WalletDetailsViewController.swift
kylehickinson: So you're hiding all the current summary rows… but then animating them hidden below to then animate them to visible after? While not accessing `summaryRows` again? I'm confused

kylehickinson: Ah, I see, you're hiding the rows that are going to be set below. I see now. Is there a reason why you don't just do 

```
// Update rows but set them hidden to be animated
self.detailsView.activityView.rows = self.summaryRows.map { $0.isHidden = true; return $0 }
```

Would probably alleviate confusion

danishjafri88: Yeah will update :)

BraveRewardsUI/Settings/Auto-Contribute/Details/AutoContributeDetailsViewController.swift
danishjafri88: Yes, the rewards db can take up some time to update the list.

PR #147, opened 2019-08-05
No Bug: Update SnapKit to 5.0.1(Swift 5 compatibility).

iccub:
This is required as we have to use the same SnapKit version on brave-ios and here.

Related ticket for upcoming Swift5 upgrade https://github.com/brave/brave-ios/pull/1154

After this one is merged. I will update SnapKit and rewards-ios dependencies in brave-ios

### DIFF
--- a/BraveRewardsUI/Common/Tables/TableViewCell.swift
+++ b/BraveRewardsUI/Common/Tables/TableViewCell.swift
@@ -83,7 +83,7 @@ class TableViewCell: UITableViewCell, TableViewReusable {
   override func updateConstraints() {
     super.updateConstraints()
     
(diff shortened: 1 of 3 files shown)


BraveRewardsUI/Common/Tables/TableViewCell.swift
```diff
@@ -83,7 +83,7 @@ class TableViewCell: UITableViewCell, TableViewReusable {
   override func updateConstraints() {
     super.updateConstraints()
     
-    label.snp_remakeConstraints {
+    label.snp.remakeConstraints {
```
iccub: fixed compiler warning

PR #148, opened 2019-08-12
Add rewards reporting to support media publishers and activity

### LINKED ISSUE #119: Replace usages of `BATBraveLedger.publisherInfoForId` method with a URL
We should also no longer be doing any conversion of URL → publisher key ourselves

This calls `GetPublisherInfo` which can return the wrong information for media publishers. We should be using `GetPublisherActivityFromUrl` which needs to be fixed

### LINKED ISSUE #137: Use new ledger observers within the brave panel UI

kylehickinson:
Fixes #119
Fixes #137 

- Panel now shows proper channels, names, etc. for supported publishers (YouTube, Vimeo, Twitch, GitHub, Twitter)
- Panel now shows proper activity for video related publishers (YouTube, Vimeo, Twitch)
- Fixes provider names not showing correctly
- Fixes grants showing probi instead of BAT
- Fixes favicons being too big for its container
- Fixes auto-contribute list item backgrounds not showing correct attention

~Library to be updated to official `master` branch Release build when https://github.com/brave/brave-core/pull/3155 is merged~ updated


BraveRewardsUI/Extensions/PublisherInfoExtensions.swift
```diff
@@ -0,0 +1,25 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import Foundation
+import BraveRewards
+
+private var AssociatedObjectHandle: UInt8 = 0
```
kylehickinson: Missed this, will delete when updating core build

BraveRewardsUI/RewardsPanelController.swift
```diff
@@ -7,12 +7,14 @@ import BraveRewards
 
 public class RewardsPanelController: PopoverNavigationController {
   
+  private var walletViewController: WalletViewController?
```
kylehickinson: Also missed this

BraveRewardsUI/Wallet/WalletViewController.swift
```diff
@@ -164,100 +206,95 @@ class WalletViewController: UIViewController, RewardsSummaryProtocol {
...
-        publisherView.checkAgainButton.isHidden = publisher.verified
-        
-        self.publisherSummaryView.setAutoContribute(enabled:
-          publisher.excluded != ExcludeFilter.filterExcluded.rawValue)
-        
-        if let percent = self.state.ledger.currentActivityInfo(withPublisherId: publisher.id)?.percent {
-          attentionView.valueLabel.text = "\(percent)%"
+
+        // Create an artificial delay so user sees something is happening
+        let diff = Date().timeIntervalSince(date)
```
kylehickinson: This is meant to be 2.5 (artificial delay) - timeIntervalSinceDate, will fix

kylehickinson:
@iccub The native libs don't do any threading themselves, so threading only changes in our layer. Background only happens from network calls, _some_ file writes and db ops.

All the DB background writes call completion on main thread: https://github.com/brave/brave-core/blob/8abf25420a8d7301f3dd4b251b62cf26cd7e5182/vendor/brave-ios/Ledger/Data/BATLedgerDatabase.mm#L17
All network completions happen on main thread: https://github.com/brave/brave-core/blob/8abf25420a8d7301f3dd4b251b62cf26cd7e5182/vendor/brave-ios/Shared/BATCommonOperations.mm#L123
File writes that happen in bg aren't involved in the callback flow

iccub:
thanks

PR #198, opened 2019-08-29
Fixes #53 by adding reachability observer

### LINKED ISSUE #53: Handle no internet and slow internet negative flows
When you launch the app, there are a number of API calls that must be made to make sure you have up to date data: 

- Publishers List
- Wallet Balance / Rates
- Wallet Properties

We need to determine how to handle these negative flows:

- User has no internet connection
- User has an extremely slow internet connection

###  _"User has no internet connection"_ options

- Show some sort of disclaimer saying their balance/data may be out of date
- Disable access to the rewards panel (however, we won't know they dont have internet connection until we try to make a call to update the data). Greyed out button (same as private mode BR button)?

###  _"User has an extremely slow internet connection"_ options

- Full panel blocked by a loader until _required data*_ needed has been collected (* - balance + wallet properties). Obviously not as nice, but hopefully won't be seen that often unless user accesses the panel before the data has been updated
- Add a loader next to the balance while its updating indicating that the shown data may not be up to date
  - **Potential caveats**: 
      - The user still has access to the panels settings, if they were to move to settings before getting BAT → USD rates, we wouldn't be able to show rates.
      - If we cache old data and show said old data, we would need to be able to indicate that said data may not be correct/out of date if the user is given freedom to use the panel with out-of-date data
      - What if your balance has not been updated and you go to tip someone? We would obviously need to block tipping while your balance is not up to date

danishjafri88:
A banner notification will be shown(cannot be closed) while there is no network


danishjafri88:
> Can we dump the Reachability dependency and just use `NWPathMonitor`? It's much less code and an official API
> 
> Edit: https://www.hackingwithswift.com/example-code/networking/how-to-check-for-internet-connectivity-using-nwpathmonitor
> 
> We use it in BATBraveAds.mm as well

Thanks for pointing out, I was not aware of this :)

PR #218, opened 2019-09-23
Fixed a couple issues with parent-child controllers.

Brandon-T:
Controllers need the proper function calls to have correct parent child relationship otherwise viewWillAppear and viewWillDisappear won't be called.

Also need to constrain the child controller's view.

### DIFF
--- a/BraveRewardsUI/Ads/AdsNotificationHandler.swift
+++ b/BraveRewardsUI/Ads/AdsNotificationHandler.swift
@@ -48,6 +48,9 @@ public class AdsNotificationHandler: BraveAdsNotificationHandler {
       presentingController.addChild(adsViewController)
       presentingController.view.addSubview(adsViewController.view)
       adsViewController.didMove(toParent: presentingController)
+      adsViewController.view.snp.makeConstraints {
+        $0.edges.equalToSuperview()
+      }
     }
     
     self.ads.reportNotificationEvent(notification.id, eventType: .viewed)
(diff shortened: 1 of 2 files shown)


BraveRewardsUI/Ads/AdsViewController.swift
```diff
@@ -297,6 +297,10 @@ extension AdsViewController {
     
     presentingController.addChild(adsViewController)
     presentingController.view.addSubview(adsViewController.view)
+		adsViewController.didMove(toParent: presentingController)
```
kylehickinson: Some sort of whitespace issue going on here (using tabs instead of spaces)

kylehickinson: Lol its still not fixed 😂

Brandon-T:
Hmm.. @kylehickinson I'm pretty sure I selected use spaces in Xcode and 2 spaces.. I just double checked.. Why it's happening I have ZERO idea.. but I had to edit it with TextEdit.. Also works if I edit it in Xcode 10.. but in 11 it completely screws up..

BraveRewardsUI/Ads/AdsViewController.swift
Brandon-T: Fixed!

BraveRewardsUI/Ads/AdsNotificationHandler.swift
```diff
@@ -47,6 +47,7 @@ public class AdsNotificationHandler: BraveAdsNotificationHandler {
     if adsViewController.parent == nil {
       presentingController.addChild(adsViewController)
       presentingController.view.addSubview(adsViewController.view)
+      adsViewController.didMove(toParent: presentingController)
```
kylehickinson: Missing setting up constraints here same as My First Ad

Issue #197, opened 2019-08-29
Implement publishers list v3 Database


kylehickinson:
Original work done here: https://github.com/brave/brave-core/pull/3217

A temporary solution was implemented for iOS to have it build for desktop due to time constraints, these changes need to be reverted for the most part and updated correctly

kylehickinson:
`verified` → `status` on `PublisherInfo` will need to be updated in brave-ios where it used to use `verified` to determine whether or not to show URL bar badge for verified publishers

PR #219, opened 2019-09-24
Ref #197: Add support for "connected" publisher status and v3 Pub list

kylehickinson:
This PR updates the library significantly far ahead

A further PR will be made in `brave-ios` to complete the changeover to publisher list v3


iccub (approved):

lib/BraveRewards.framework/Headers/ledger.mojom.objc.h
```diff
@@ -61,11 +61,27 @@ typedef NS_ENUM(NSInteger, BATResult) {
...
   BATResultRecurringTableEmpty = 23,
   BATResultExpiredToken = 24,
+  BATResultBatNotAllowed = 25,
+  BATResultAlreadyExists = 26,
 } NS_SWIFT_NAME(Result);
 
 
+typedef NS_ENUM(NSInteger, BATPublisherStatus) {
+  BATPublisherStatusNotVerified = 0,
+  BATPublisherStatusConnected = 1,
```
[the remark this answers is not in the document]
kylehickinson: It means that the publisher has connected a wallet to the account but not yet verified

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
```diff
@@ -264,7 +264,11 @@ extension TipsDetailViewController {
   fileprivate func getTipsThisMonth() -> BalanceReportInfo {
     let month = Date().currentMonthNumber
     let year = Date().currentYear
-    return state.ledger.balanceReport(for: ActivityMonth(rawValue: month) ?? .any, year: Int32(year))
+    var report = BalanceReportInfo()
+    state.ledger.balanceReport(for: ActivityMonth(rawValue: month) ?? .any, year: Int32(year)) {
+      if let balance = $0 { report = balance }
```
Brandon-T: This function is async.. but it returns the report before the function completes?
Is it truly non-async but has a callback instead of return-statement?

lib/BraveRewards.framework/Headers/ledger.mojom.objc.h
```diff
@@ -61,11 +61,27 @@ typedef NS_ENUM(NSInteger, BATResult) {
...
+  BATPublisherStatusVerified = 2,
+} NS_SWIFT_NAME(PublisherStatus);
 
-@class BATContributionInfo, BATPublisherInfo, BATPublisherBanner, BATPendingContribution, BATPendingContributionInfo, BATVisitData, BATGrant, BATWalletProperties, BATBalance, BATAutoContributeProps, BATMediaEventInfo, BATExternalWallet, BATBalanceReportInfo, BATActivityInfoFilterOrderPair, BATActivityInfoFilter, BATReconcileInfo, BATRewardsInternalsInfo;
+
+typedef NS_ENUM(NSInteger, BATRewardsCategory) {
+  BATRewardsCategoryAutoContribute = 2,
+  BATRewardsCategoryOneTimeTip = 8,
+  BATRewardsCategoryRecurringTip = 16,
```
Brandon-T: Would be much better to use bit-shifting to be more explicit that the enum values can be "OR'd" than hardcoding the result of it.

IE:

```
1 << 1  = 2
1 << 2 = 4
1 << 3  = 8
1 << 4 = 12
1 << 5  = 16
```

and so on.

Brandon-T (changes requested):

BraveRewardsUI/Settings/Tips/Details/TipsDetailViewController.swift
kylehickinson: It is non-escaping, executed serially. They just changed the API on native libs side for memory reasons I believe?

lib/BraveRewards.framework/Headers/ledger.mojom.objc.h
kylehickinson: This is autogenerated based on this mojom file: https://github.com/brave/brave-core/blob/4d2e01bc756cc5c76a8444509c4182cbc47e3c35/vendor/bat-native-ledger/include/bat/ledger/public/interfaces/ledger.mojom#L248

We aren't defining these values. I can only assume they started out like that and options were removed or something

Brandon-T (approved):

danishjafri88 (approved):

Brandon-T (approved):

Issue #229, opened 2019-10-01
Fix colors/themeing in rewards panel (iOS 12 & iOS 13)


kylehickinson:
We don't plan on supporting themes for Brave Rewards at the moment, but nav bars/background colors/etc. are appearing dark when run in dark mode regardless.

Should be showing only light mode regardless of system pref/theme selection

This affects both iOS 12 and iOS 13

iOS 12 has entirely incorrect text colors among other things, iOS 13 has incorrect table colors:

![IMG_4356](https://user-images.githubusercontent.com/529104/65997802-0f8ad800-e468-11e9-8653-64111e1e43c3.png)

PR #232, opened 2019-10-03
Ref #229: Always use light theme for Rewards UI on iOS 13.

iccub:
This fix makes all rewards UI on iOS 13 to look good.

fixing iOS 12 will be harder, if not impossible without moving the codebase over to brave-ios


kylehickinson:
OK cool, after onboarding and publisher list merges, I'll move the code in this repo to brave-ios

